PR #34, opened 2022-02-08
Make Fleks multiplatform compatible

jobe-m:
- Adapt registering of systems in World

Instantiation of generic classes is not possible in Kotlin
multiplatform. Thus, instead of instantiating a generic class in the
SystemService the system function gets a factory method for calling
the constructor of the system.

- Introduce adding of component and its listener in world configuration

The component listener will be added together with registering the
component in the world configuration. That was done because when
registering the component listener the base type of the component is
needed. But without java reflections it is not possible to detect the type
of component listeners. Thus, adding component and listener objects
together is the most elegant workaround for now - I hope.

- Adapted readme to reflect changed API

- Adapted Fleks benchmark

~~Currently registering of systems is not working in the benchmark
`fleks.kt`. In my Korge example it is working that way. It could be that this is
due to some multiplatform configuration in Korge. But currently this is
beyond my understanding. :(~~

Stupid mistake. Now the benchmark is working. Only the complex one is
failing due to an exeption.

- Renaming some variables

I changed `cmp` to `comp` because especially at the beginning I read e.g. `cmpMapper`
as "compareMapper" and not "componentMapper" which was a bit confusing.


ReadMe.md
```diff
@@ -70,7 +70,7 @@ implementation 'io.github.quillraven.fleks:Fleks:1.0-RC3'
 #### Gradle (Kotlin)
 
 ```kotlin
-implementation("io.github.quillraven.fleks:Fleks:1.0-RC3")
+implementation("io.github.quillraven.fleks:Fleks:1.0-MPP")
```
Quillraven, line 73: Is it MPP or KMM? If it is MPP then I will change the branch name ;)

jobe-m: I just looked it up and the correct name is KMP for Kotlin MultiPlatform. KMM  stands for Kotlin Multiplatform Mobile. So would be only a special part of it;)

Quillraven:
Again, thank you a lot for the PR! I will have a look in more detail by end of this week and try to run the benchmarks on my machine and let you know about the results.

Four things:
- I will add you as a contributor to this project or is this done automatically by github when your PR gets merged?
- I will add you at the top of the ReadMe as well. At this place I will then mention the multi platform branch and the separate version on maven for it. So that people know about it and can use it e.g. in their Korge games.
- I will release a maven version of the multi platform branch by end of this week
- How is the multiple dependency for the same class working? I saw you removed that part from the ReadMe and that the dependency map now uses `KClass` instead of `String`. So I am wondering how it works :D

ReadMe.md
Quillraven: Okay, I changed it to kmp. Please also change it then in the ReadMe and buil.gradle.kts version ;)

jobe-m:
Sounds good! Thanks:)
No hurry, take your time.
You don't need to add me as contributor technically in GitHub. I can always send you PRs from the forked repo.
Regarding multiple dependencies of the same type: this is actually not working/I couldn't get it working since the used annotations are a java feature. So I removed that section from the readme.

ReadMe.md
```diff
@@ -94,15 +97,20 @@ to configure it accordingly:
...
     entityCapacity = 1000
 
-    system<MoveSystem>()
-    system<PhysicSystem>()
+    system(::MoveSystem)
+    system(::PhysicSystem)
+
+    component(::Position)
+    component(::Physic)
+    component(::Box2dComponent, ::Box2dComponentListener)
```
Quillraven, line 113: I like that change a lot because in my opinion a listener is always for one specific component and registering it directly with the Component is a cool idea. I will take that over to the master as well.

ReadMe.md
```diff
@@ -119,12 +127,12 @@ called. It is a made up example of a Day-Night-Cycle system which switches betwe
...
 
 ```Kotlin
-class DayNightSystem(
-    private val eventMgr: EventManager
-) : IntervalSystem() {
+class DayNightSystem() : IntervalSystem() {
     private var currentTime = 0f
     private var isDay = false
 
+    private val eventMgr: EventManager = Inject.dependency()
```
Quillraven, line 134: Out of curiosity as I saw something similar in other libraries: would it be possible to replace `Inject.dependency()` or `Inject.componentMapper()` by a `by inject()` statement? Similar to Kotlin's `by lazy()`?

ReadMe.md
```diff
@@ -47,7 +47,10 @@ then use [Artemis-odb](https://github.com/junkdog/artemis-odb) or [Ashley](https
 
 ## Current Status
 
-Release Candidate version 1.0-RC3 is available on maven central since 30-Jan-2022. Please feel free to contribute to the
+This branch contains a special Kotlin Multiplatform compatible verion of Fleks. It has a slightly different API when compared with the normal 1.0 release.
+But still similar enough that it should not be much work to switch between (normal) release 1.0 and release 1.0-KMP.
+
+Release version 1.0-KMP-RC3 is available on maven central since XX-YYY-2022. Please feel free to contribute to the
```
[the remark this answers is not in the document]
jobe-m: Yes, of course :)

src/main/kotlin/com/github/quillraven/fleks/collection/bitArray.kt
```diff
@@ -141,7 +141,6 @@ class BitArray(
 
     override fun equals(other: Any?): Boolean {
         if (this === other) return true
-        if (javaClass != other?.javaClass) return false
```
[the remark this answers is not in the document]
jobe-m: I added it also now to the branch.

jobe-m:
> Regarding multiple dependencies of the same type: this is actually not working/I couldn't get it working since the used annotations are a java feature. So I removed that section from the readme.

While reading about delegates I stumbled over KPropery and realized that it can be used to get the name of the property. With that we might be able to implement the multiple dependency injection if the same type also without annotations. I will check :)

Quillraven:
> > Regarding multiple dependencies of the same type: this is actually not working/I couldn't get it working since the used annotations are a java feature. So I removed that section from the readme.
> 
> While reading about delegates I stumbled over KPropery and realized that it can be used to get the name of the property. With that we might be able to implement the multiple dependency injection of the same type also without annotations. I will check :)

Sounds promising! Then let's wait with the RC1 release. Maybe you can add that support that would be great! Then everything is still covered from my understanding only the API is slightly different :)

ReadMe.md
Quillraven: I agree with you. If this adds the overhead of an extra object/property and evaluates it every time then let's not do that for performance reasons ;)

Thanks for checking!

jobe-m:
I now realized that KProperty is only usable when delegating access to a property. But again I don't want that for injected objects because of "delayed/indirect" access.
But I have another idea. Just using a String to specify the object type when injecting should do the trick :)

jobe-m:
I changed the map keys of storing the dependency objects from KClass<*> back to String. Now it is also possible to specify the "type name" (key in the map of dependencies) in the inject() and the Inject.dependency() functions.
For the external usable API the only difference is now that Inject.dependency() can also have a parameter with the type name.
So, now the last part is also working again. :)

Quillraven:
Great job! I will then prepare the release of mavenCentral by end of this week and let you know when it is done.

Thank you!

PR #36, opened 2022-02-26
Fix unit tests and update sources

jobe-m:
This commit fixes the unit test run. It updates the unit tests to the
changed API which was needed to remove the dependency to java
annotations, etc. Some minor test cases are not needed anymore because
annotations are not used to create objects. Instead objects are created
using constructor functions. That makes some test cases obsolete. Also
some test cases were changed to check the adapted exeption objects.

Some minor changes in the Fleks sources were done due to cleanup code
which was before used by annotations. Also the registering of component
listeners was made more robust by defining the expected type T. That was
found while getting the unit tests running :)

### DIFF
--- a/src/main/kotlin/com/github/quillraven/fleks/world.kt
+++ b/src/main/kotlin/com/github/quillraven/fleks/world.kt
@@ -81,10 +81,9 @@ class WorldConfiguration {
      * @param compFactory the constructor method for creating the component.
      * @param listenerFactory the constructor method for creating the component listener.
      * @throws [FleksComponentAlreadyAddedException] if the component was already added before.
-     * @throws [FleksComponentListenerAlreadyAddedException] if the listener was already added before.
      * @throws [FleksInjectableTypeHasNoName] if the dependency type has no T::class.simpleName.
      */
-    inline fun <reified T : Any> component(noinline compFactory: () -> T, noinline listenerFactory: (() -> ComponentListener<*>)? = null) {
+    inline fun <reified T : Any> component(noinline compFactory: () -> T, noinline listenerFactory: (() -> ComponentListener<T>)? = null) {
         val compType = T::class.simpleName ?: throw FleksInjectableTypeHasNoName(T::class)
 
         if (compType in componentFactory) {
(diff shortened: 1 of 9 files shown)


src/main/kotlin/com/github/quillraven/fleks/world.kt
```diff
@@ -81,20 +81,17 @@ class WorldConfiguration {
...
 
         if (compType in componentFactory) {
             throw FleksComponentAlreadyAddedException(compType)
         }
         componentFactory[compType] = compFactory
         if (listenerFactory != null) {
-            if (compType in compListenerFactory) {
-                throw FleksComponentListenerAlreadyAddedException(compType)
-            }
+            // No need to check compType again in compListenerFactory - it is already guarded with check in componentFactory
```
jobe-m, line 94: While writing/checking the unit test I realized that this check is not needed since the listener type cannot be different to the component type. Otherwise there will be already a compile error. This is nice since the user cannot do anything wrong when registering the component together with the listener object.

src/main/kotlin/com/github/quillraven/fleks/world.kt
```diff
@@ -142,6 +139,11 @@ class World(
         componentService = ComponentService(worldCfg.componentFactory)
         entityService = EntityService(worldCfg.entityCapacity, componentService)
         val injectables = worldCfg.injectables
+
+        // Add world to inject object so that component listeners can get it form injectables, too
+        // Set "used" to true to make this injectable not mandatory
+        injectables["World"] = Injectable(this, true)
```
jobe-m, line 145: I added that because in my example (which I write for Korge) I need the world object in a component listener. Since the listener is just an interface there is now the possibility to add a dependency object for the "World" in the final component listener which is injected automatically :)

src/test/kotlin/com/github/quillraven/fleks/ComponentTest.kt
```diff
@@ -142,26 +155,17 @@ internal class ComponentTest {
...
+        val cmpService = ComponentService(componentFactory)
         val expected = cmpService.mapper<ComponentTestComponent>()
 
         val actual = cmpService.mapper<ComponentTestComponent>()
 
         assertSame(expected, actual)
     }
 
-    @Test
-    fun `cannot create mapper for component without no-args constructor`() {
```
jobe-m, line 154: This test case is obsolete with the constructor functions from the componentFactory.

Quillraven:
Great stuff, thanks a lot! I have some time tomorrow afternoon where I will then finally release the kmp version :)

PR #37, opened 2022-03-19
Update with changes done in korge-fleks

jobe-m:
New feature was added which makes it possible to mark an injectable as
"used". Thus Fleks will not throw exception if the injectable is not
used directly at creation of system and listener objects. Ie. that makes
it possible to use the injected objects during init functions. Default
for use is still "false" as before.

### DIFF
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/collection/bitArray.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/collection/bitArray.kt
@@ -140,10 +140,9 @@ class BitArray(
     }
 
     override fun equals(other: Any?): Boolean {
-        if (other == null) return false
         if (this === other) return true
+        if (other !is BitArray) return false
 
-        other as BitArray
         val otherBits = other.bits
 
         val commonWords: Int = min(bits.size, otherBits.size)
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/system.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/system.kt
@@ -149,13 +149,13 @@ abstract class IteratingSystem(
 ) : IntervalSystem(interval, enabled) {
     /**
      * Returns the [family][Family] of this system.
-     * This reference gets updated by the [SystemService] when the system gets created via reflection.
+     * This reference gets updated by the [SystemService] when the system gets created via the SystemFactory.
      */
     internal lateinit var family: Family
 
     /**
      * Returns the [entityService][EntityService] of this system.
-     * This reference gets updated by the [SystemService] when the system gets created via reflection.
+     * This reference gets updated by the [SystemService] when the system gets created via the SystemFactory.
      */
     @PublishedApi
     internal lateinit var entityService: EntityService
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/world.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/world.kt
@@ -52,26 +52,34 @@ class WorldConfiguration {
     /**
      * Adds the specified [dependency] under the given [type] which can then be injected to any [IntervalSystem] or [ComponentListener].
      *
+     * @param type is the name of the dependency which is used to access it in systems and listeners. This is especially useful if two or more
+     *             dependency objects of the same type shall be injected.
+     * @param dependency object which shall be injected to systems and listeners of the Fleks ECS.
+     * @param used this will set the injected dependency to [used] internally. Default is false. If set to true then Fleks will not
+     *             complain if the dependency is not used by any system or listener on their creation time.
      * @throws [FleksInjectableAlreadyAddedException] if the dependency was already added before.
      */
-    fun <T : Any> inject(type: String, dependency: T) {
+    fun <T : Any> inject(type: String, dependency: T, used: Boolean = false) {
         if (type in injectables) {
             throw FleksInjectableAlreadyAddedException(type)
         }
 
-        injectables[type] = Injectable(dependency)
+        injectables[type] = Injectable(dependency, used)
     }
 
     /**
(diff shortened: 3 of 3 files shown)


Quillraven:
I changed the target branch  to `next` which contains some version updates and also a bugfix for the benchmarks. The fleks benchmark doesn't work in the `kmp` branch because of wrong system creation arguments :D Seems like it got changed from `allOf` to `allOfComponents`, etc.. 

Also, I am wondering but didn't you have to modify some tests as well because fleks was using JUnit5 but Korge uses JUnit4?

By the way - thank you for your commitment and support with Fleks!

jobe-m:
Ah, right. I changed the allOff arguments and removed the family classes. It was kind of obsolete without annotations;)
I guess here in Fleks we can keep junit5. That is why I did not take those changes over.

Quillraven:
@jobe-m : I think I was able to convert the jvm project to a multiplatform project with currently only a jvm target. JS/native would theoretically work as well (code is commented out in `build.gradle.kts`) but some tests fail because e.g. Javascript has no IndexOutOfBoundsException. You can access invalid indices and I guess it then simply returns undefined ;)

For native there were other errors that I didn't look into yet. For JS I can imagine having separate unit tests for JVM/JS OR validate the out of bounds issues in Fleks and throw a custom exception but that will not align with the standards of those languages. I think different unit tests are a better way, what do you think?

The branch is called `#38_gradle-kmp-setup`. It is not final yet as I want to bring in js and native target as well. When you remove the `signing` plugin and the `signing {...}` block then you can execute the `publishToMavenLocal` task to create a local maven artifact.

I tried that with Dinoleon, the example game, and it looks like it works there including showing the comments. I removed some dokka custom tasks which seem to be not necessary anymore with the multiplatform setup. 

I would appreciate it, if you can test that with KorGE. I think you should be able now to add Fleks as a normal "commons" dependency and it should hopefully work. At least for the jvm target at the moment.

For this PR I want o wait for the final gradle setup and then I will merge it :)

Quillraven:
Okay, was now able to add JS and native target as well. For JS, as mentioned, I split up some of the tests. And for native everything was actually working when adding `@ThreadLocal` to the `Inject` object. Otherwise, it will fail because it is a frozen object that gets mutated.

Can you rebase your branch to this one please and try out that your changes are still working?

Sorry for the inconvenience but this should be a one-time thingy when we switch now to KMP for the future.

jobe-m:
Great, sounds good! I will check it when I get some time. Probably the next days...

jobe-m:
I could build locally this rebased branch and run publishToMavenLocal successfully.
But I don't know now I add it to the gradle script of my Korge game project.

fleks-in-jobesLegacy

Line 18 is what I added so that Fleks should be taken from maven local repo.
Line 28 is commented out. That was taking Fleks from korge-fleks module in Korge.

Quillraven:
What a weird topic and I cannot find any good documentation. But thanks for trying! Found this thread but there was no real answer: https://forum.korge.org/topic/14/adding-external-libraries/2

So I fiddled around myself:
- install Korge plugin for Intellij
- create new project
- update build.gradle.kts file
```kotlin
import com.soywiz.korge.gradle.KorgeGradlePlugin
import com.soywiz.korge.gradle.korge

buildscript {
    val korgePluginVersion: String by project

    repositories {
        mavenLocal()
        mavenCentral()
        google()
        maven { url = uri("https://plugins.gradle.org/m2/") }
    }
    dependencies {
        classpath("com.soywiz.korlibs.korge.plugins:korge-gradle-plugin:$korgePluginVersion")
    }
}

apply<KorgeGradlePlugin>()

korge {
    id = "com.example.example"
    supportBox2d()
    dependencyMulti("io.github.quillraven.fleks:Fleks:2.0-RC1", registerPlugin = false)
// To enable all targets at once

    //targetAll()

// To enable targets based on properties/environment variables
    //targetDefault()

// To selectively enable targets

    targetJvm()
    //targetJs()
    //targetDesktop()
    targetIos()
    targetAndroidIndirect() // targetAndroidDirect()
}
```
- the important part is the `dependencyMulti` line which adds Fleks. I got that by looking at what `supportBox2d` does internally
- I had to deactivate `targetJs` and `targetDesktop` because it complains that it cannot resolve Fleks for those targets. I dont' understand yet why but I guess those targets are not properly published yet by Fleks???
- updated the `main.kts` file
```kotlin
suspend fun main() = Korge(width = 512, height = 512, bgcolor = Colors["#2b2b2b"]) {
    val minDegrees = (-16).degrees
    val maxDegrees = (+16).degrees

    val image = image(resourcesVfs["korge.png"].readBitmap()) {
        rotation = maxDegrees
        anchor(.5, .5)
        scale(.8)
        position(256, 256)
    }

    class TestSystem : IntervalSystem(interval = Fixed(1f)) {
        override fun onTick() {
            println("tick")
        }
    }

    val world = World {
        system(::TestSystem)
    }

    while (true) {
        image.tween(image::rotation[minDegrees], time = 1.seconds, easing = Easing.EASE_IN_OUT)
        image.tween(image::rotation[maxDegrees], time = 1.seconds, easing = Easing.EASE_IN_OUT)
        world.update(1f)
    }
}
```
- be astonished at the "tick" log every 3 seconds or so. I guuess the tween stuff is somehow blocking???

----

Will need to invest more time to understand why JS and desktop targets are not working. I really assume that they are not properly published somehow but I don't know why at the moment. Any help is appreciated! :)

jobe-m:
Thanks for looking into this problem. I just tried it out and it works also for me :)
And you are right with adding the lib with dependencyMulti(). I now remember that I have seen it also in the past when I added the "supportFleks()" function into the korge-gradle-plugin. I just did not remember it in the first run.

Runing my Korge game with the jvm target and Fleks from maven local is now working. It also compiles for linux native but the exe crashes on startup. But I have the same error when using Fleks form Korge-fleks module. The Js target works also fine :)

Issue #40, opened 2022-05-07
Entity creation in a system init block does not work properly


Quillraven:
A user in Discord (Hafu) contacted me that he has a problem creating an entity in the `onInit` block of a system.
This function got removed in the last update but I think the problem still remains and makes sense to me because families are created together with their system and therefore if a system creates an entity in its init block then not all families will be notified at that time.

One solution could be to iterate over all entities after system creation and trigger an update for each family to correctly update them. This is a one time operation during world creation and should not be a big problem.

We also need to check the `ComponentListener` that they also get notified correctly during world creation.

Quillraven:
PR is open. I will merge it tomorrow or Thursday and will publish a new version.

Quillraven:
part of 1.1-JVM version which gets published right now!

PR #41, opened 2022-05-10
fix issue that families are not updated correctly during system creation

Quillraven:
Fix for #40 
this is also relevant for the upcoming KMP version

----

A different solution would be to first create the families in the `SystemService` and afterwards the systems. However, this requires to iterate over the `systemTypes` parameter twice and also find the family again for each system in the `allFamilies` list.

I did not like that a lot that's why I added a separate iteration over all families and entities afterwards. I guess entity creation in a system init block is an exceptional case so the iteration count should be minimal.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -127,6 +127,12 @@ publishing {
     }
 }
 
+// only sign if version is not a SNAPSHOT release.
+// this makes it easier to publish to mavenLocal and test the packed version.
+tasks.withType<Sign>().configureEach {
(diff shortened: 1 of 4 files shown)


Quillraven:
Refactored it a little bit by introducing a `notifyAll` function to the `EntityService`. This is cleaner as it will also consider other `EntityListener` that might be registered.

Also, signing is now only happening for non-SNAPSHOT releases. That way it is easier to publish to mavenLocal for testing. Just change the version to something that ends with SNAPSHOT ;)

PR #50, opened 2022-06-02
Merge master branch content into next branch

jobe-m:
The content on the next branch should not be the same as for release 1.2-JVM.
After some testing we could publish release 1.2-KMP.

I will also try to use directly this version in my stand alone Korge game. If that works I could drop Korge-fleks again.

### DIFF
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -1,7 +1,7 @@
 # Fleks
 
 [![MIT license](https://img.shields.io/badge/License-MIT-blue.svg)](https://github.com/Quillraven/Fleks/blob/master/LICENSE)
-[![Maven](https://img.shields.io/badge/Maven-1.1--KMP-success.svg)](https://search.maven.org/artifact/io.github.quillraven.fleks/Fleks/1.1-KMP/jar)
+[![Maven](https://img.shields.io/badge/Maven-1.2--KMP-success.svg)](https://search.maven.org/artifact/io.github.quillraven.fleks/Fleks/1.2-KMP/jar)
 
 [![Build KMP](https://img.shields.io/github/workflow/status/quillraven/fleks/Build/next?event=push&label=Build%20next)](https://github.com/Quillraven/fleks/actions)
 [![Kotlin](https://img.shields.io/badge/Kotlin-1.6.21-red.svg)](http://kotlinlang.org/)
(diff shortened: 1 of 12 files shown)


jobe-m:
Argh, I just run the commonTest locally. It looks like the CURRENT_WORLD companion object is making trouble on native. It is not initialized in some unit tests.
I remember also some trouble in the same direction with the Injection object in Korge-fleks. It was needed to be labeled as @threadLocal...

Quillraven:
Wow you are fast! I wagely remember that I had a similar issue in the KMP environment and I think you are right with  @ThreadLocal.

But please let's wait for 1.3-JVM because there are also two bugfixes in it that are important when it comes to nested loops with families and also regarding notification of ComponentListener.

I try to finalize it by today :)

Thank you, as always, for your support!

jobe-m:
No problem, I just had some the yesterday :)
The next days I won't have time due to preparations for vacation ;-D
Feel free to update here with the release 1.3 content. I kept the state of the other older versions in the commits in case we want to go back also in the KMP/next branch.

Quillraven:
I try to have a look over the weekend, if I can find some time. Enjoy your holiday! :)

jobe-m:
I finally have Internet again in France. We were driving cross Switzerland and had no roaming without high costs ;)
Take your time 👍

Quillraven:
This PR is now broken because of me :(

I did the version updates (gradle, Kotlin), merged the ComponentListener notification fix and refactored the dependency setting/clearing as discussed in the other PR to support different dependencies for each world. TBH I think that already worked before but now it clears the dependencies of the Inject object at the end of the world creation to let the GC clean them up.

I also updated the ReadMe. It contained some outdated code snippets as well. It mentions already the new features but they are not yet included.

I try to do that tomorrow. Your PR and diffs help with that a lot ;)

jobe-m:
That, sound good. Let me know if I can help in resolving the merge conflicts. When I am back at my laptop I can check it locally from maven.

Quillraven:
@jobe-m / @Kietyo: I updated the `next` branch with all the changes of the JVM branch including the new `FamilyListener` stuff. It would be great if you can test this branch in your games/examples to verify it.

You can publish it to your local maven repository via the `publishToMavenLocal` gradle task. Since the version is defined as 1.3-KMP-SNAPSHOT you don't need to sign it and it should just work.

In your KorGE application you need to add it as a dependency like this:

```kotlin
dependencyMulti("io.github.quillraven.fleks:Fleks:1.3-KMP-SNAPSHOT", registerPlugin = false)
```

That should be it! I will also try that this week with a fresh new KorGE project but since I do not have a lot of experience with KorGE, I'd appreciate your input!

When everything is fine, I will merge those changes into the kmp branch and release a new 1.3-KMP version to maven.

Quillraven:
> I quickly checked the state of the next branch in my game and have a problem with dependency injection. It looks like the injector does not work as before. I need to check further. Did you change there something?

Not the injector itself but the dependencies/mappers are now set and cleared in the world constructor. Since the tests passed I thought it was fine but obviously it isn't. I will make a PR to KMP branch then it is easier to see the diffs.

Quillraven:
I will close this PR. Let's continue the discussion in #53

PR #51, opened 2022-06-03
Deprecate global inject object for world local injections.

Kietyo:
The idea is that each world should have their own injection space. Currently, every world that's created overwrites the previous world's injections. So you can't have 2 worlds with different injections.

This is useful for simulation/multiplayer type games where each player has their own world.

A real life example of this is that for my game, I am replicating a starcraft 2 arcade map called "Entropy TD" where each player is competing against each other with their own mazes. Each player has their own world with their own instances of injectables.

https://sc2arcade.com/map/2/148730/

### DIFF
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/Injections.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/Injections.kt
@@ -1,0 +1,45 @@
+package com.github.quillraven.fleks
+
+/**
+ * An [injector][Injections] which is used to inject objects from outside the [IntervalSystem].
+ *
+ * @throws [FleksSystemDependencyInjectException] if the Injector does not contain an entry
+ * for the given type in its internal map.
+ * @throws [FleksSystemComponentInjectException] if the Injector does not contain a component mapper
+ * for the given type in its internal map.
+ * @throws [FleksInjectableTypeHasNoName] if the dependency type has no T::class.simpleName.
+ */
+data class Injections(
+    @PublishedApi
+    internal val injectObjects: Map<String, Injectable> = mapOf(),
+    @PublishedApi
+    internal val mapperObjects: Map<String, ComponentMapper<*>> = mapOf()
+
+) {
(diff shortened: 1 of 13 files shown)


Quillraven:
In my opinion injectables should not be kept after world creation. They are only needed to start up your world by creating systems, ComponentListener and in the near future FamilyListener. Maybe EntityComparator as well. But afterwards they are not necessary anymore and the GC should be allowed to clean them up.

Therefore, I suggest to not store them as part of those objects. 

As a proposal: can we not just initialize the existing `Inject` object at the beginning of the world's constructor and then clear it afterwards? That way your use-case should work as well and we don't need to keep the injectable maps in memory.

Something like:
```diff
// world.kt

init {
    val worldCfg = WorldConfiguration().apply(cfg)
    componentService = ComponentService(worldCfg.componentFactory)
    entityService = EntityService(worldCfg.entityCapacity, componentService)
    val injectables = worldCfg.injectables
    
    // Add world to inject object so that component listeners can get it form injectables, too
    // Set "used" to true to make this injectable not mandatory
    injectables["World"] = Injectable(this, true)

+  Inject.injectObjects = injectables
+  Inject.mapperObjects = compService.mappers
    
    // create and register ComponentListener
    worldCfg.compListenerFactory.forEach {
        val compType = it.key
        val listener = it.value.invoke()
        val mapper = componentService.mapper(compType)
        mapper.addComponentListenerInternal(listener)
    }

    systemService = SystemService(this, worldCfg.systemFactory, injectables)
    
    // verify that there are no unused injectables
    val unusedInjectables = injectables.filterValues { !it.used }.map { it.value.injObj::class }
    if (unusedInjectables.isNotEmpty()) {
        throw FleksUnusedInjectablesException(unusedInjectables)
    }

+  Inject.injectObjects = emptyMap()
+  Inject.mapperObjects = emptyMap()
}
```

Or am I missing something?

Kietyo:
Thanks, I think that will work for now.

PR #53, opened 2022-06-07
Update KMP with changes of 1.1-JVM to 1.3-JVM

Quillraven:
Changelog:
- [x] BUGFIX: Entities created inside a system's constructor are now properly added to their family and also `EntityListener` get notified correctly in this case
- [x] BUGFIX: `ComponentListener` did not get notified when entities are created in a system's `init` block
- [x] UPDATE: change dokka from 1.6.10 to 1.6.21
- [x] UPDATE: the `used`  parameter of an injectable is removed because injectables are now properly resolved and cleared once the world is created
- [x] NEW: make world a default dependency for the dependency injection mechanism. That way it is now possible to easily pass the world to a `ComponentListener`
- [x] NEW: add getter for world's `systems`
- [x] NEW: add `configureEntity` to world. This makes it now possible to modify an entity outside of a system as requested in the comments section of my YouTube channel
- [x] NEW: add support to create families out of a system via the world. This is the biggest change of this version and adds new flexibility on how to handle and iterate over entities with specific component configurations
- [x] NEW: users have now access to the `family` of an `IteratingSystem` 
- [x] NEW: `family` now also contains a `configureEntity` function which allows users to modify an entity outside of a system in a convenient way
- [x] NEW: a new `FamilyListener` support is added which allows users to react on when an entity is added to, or removed from a `family`    
- [x] NEW: new DSL to create a world. Use `components` , `systems`, `injectables` or `families` block to define your world. For more details, refer to the ReadMe
- [x] NEW: the KMP setup was updated to a proper multiplatform gradle setup. That way it is now possible to add Fleks to a KorGE dependency via following line in the `korge` block
    ```kotlin
    dependencyMulti("io.github.quillraven.fleks:Fleks:1.3-KMP", registerPlugin = false)
    ```

----

This PR contains all the changes that were missing in the KMP branch that got added in the 1.1-JVM, 1.2-JVM and 1.3-JVM version.

In addition it changes the gradle setup to be a real KMP project. Most of the code is in `commonMain` but there are some testcases in other source sets because when it comes to exceptions then the different backends are behaving differently.

Not sure if this is the best solution to do it but for now let's roll with it ;) As an example, in JS there is no `IndexOutOfBoundsException` when accessing an invalid array position. It won't show any error but the result is most likely `undefined`.

It also contains some changes that were done in this PR: https://github.com/Quillraven/Fleks/pull/37

----

With the new setup it should be now possible to add Fleks as a normal dependency to KorGE like this:

```kotlin
dependencyMulti("io.github.quillraven.fleks:Fleks:1.3-KMP", registerPlugin = false)
```


jobe-m:
I didn't had time yet to check further. Maybe I just need to adapt a bit in my code. Will let you know.

src/commonMain/kotlin/com/github/quillraven/fleks/world.kt
```diff
@@ -0,0 +1,411 @@
...
+        // verify that there are no unused injectables
+        val unusedInjectables = injectables.filterValues { !it.used }.map { it.value.injObj::class }
+        if (unusedInjectables.isNotEmpty()) {
+            throw FleksUnusedInjectablesException(unusedInjectables)
+        }
+
+        // clear dependencies at the end because they are no longer necessary,
+        // and we don't want to keep a reference to them
+        Inject.injectObjects = EMPTY_INJECTIONS
+        Inject.mapperObjects = EMPTY_MAPPERS
```
jobe-m, line 239: I found my problem. I was accessing the injectObjects property after creation of my specific systems and component listeners. This was not intended. Thus I will change my code. No changes on your side needed.

jobe-m:
BTW from my side we can delete the kmp branch and use the next branch for further "KMP" releases.

Quillraven:
I will try to setup a KorGE project tomorrow to also verify some things but good to know that you found your issue! :)

About the branches: I am not sure yet ... I think we can keep master for JVM and kmp for KMP at the moment. They will always reflect the latest developments.
In addition we have the release tags and of course feature branches for developments.

I will delete `next` after the merge,

jobe-m:
Ok, that's also fine with me. Will you add the DSL config for the world to kmp branch then?

Quillraven:
> Ok, that's also fine with me. Will you add the DSL config for the world to kmp branch then?

I can do that. Next week is a short one in Austria, so I will have some time to work on that 🙂

Quillraven:
@jobe-m : I removed the `used` parameter and added the DSL for the world configuration. Unfortunately, for whatever reason, I cannot make it work to add it as a KorGE dependency.

From what I remember I simply published to mavenLocal and added Fleks as a dependency like this:

```kotlin
korge {
	id = "com.example.example"
	dependencyMulti("io.github.quillraven.fleks:Fleks:1.3-KMP-SNAPSHOT", registerPlugin = false)
// To enable all targets at once

	//targetAll()

// To enable targets based on properties/environment variables
	//targetDefault()

// To selectively enable targets

	targetJvm()
	targetJs()
	targetDesktop()
	targetIos()
	targetAndroidIndirect() // targetAndroidDirect()
}
```

This was working before but now for whatever reason I get errors like this:

```
korgetest2:jvmTest: Could not resolve io.github.quillraven.fleks:Fleks:1.3-KMP-SNAPSHOT.
Required by:
    project :

Possible solution:
 - Declare repository providing the artifact, see the documentation at https://docs.gradle.org/current/userguide/declaring_repositories.html
```

No idea what that means or what it is trying to tell me :D The artifact is for sure available because I can add it without any problems to a normal gradle setup. Don't know what is so special about KorGE :D 

Do you have any idea?

soywiz:
@Quillraven usually when an artifact is not found I believe here:

```
Possible solution:
 - Declare repository providing the artifact,
```

It displays a list of artifacts where it searched. If it didn't search, maybe not repositories are defined?

Also if that's an SNAPSHOT, it couldn't be published to mavenCentral, but maybe to the snapshot repository that is not included automatically for security.
It you uploaded to central snapshots  you can try something like:

```kotlin
repositories {
  maven { url = uri("https://oss.sonatype.org/content/repositories/snapshots") }
}
```

jobe-m:
Did you publish all artifacts to mavenLocal? Not only one like the JVM target of Fleks. I remember that was not working. Otherwise your gradle setup looks ok. The setup of my game should be identical.
Next time when I have my laptop reachable I will check this PR and the DSL config.
We are currently on the way back home from vacation. Many greetings from a camping side at the lake Geneva. :)

jobe-m:
> Also if that's an SNAPSHOT, it couldn't be published to mavenCentral

I checked publishing Fleks 1.3-KMP-SNAPSHOT two weeks ago to mavenLocal and using it in my game was possible. So the version name should be ok when used locally.

soywiz:
So this is about it not finding the version in the `.m2` folder? I believe I misunderstood then. If that's so I have no idea. Then it should include a:

```kotlin
repositories {
  mavenLocal()
  // ...
}
```

Quillraven:
Exactly, it seems like the KorGE project cannot load it properly. I can add the dependency just fine to a LibGDX project. So I guess either my publication is wrong (but it worked in March) or something else has changed since March.

In my .m2 folder I see four Fleks folders:
image

They all contain a 1.3-KMP-SNAPSHOT folder:
image

With imo correct artifact files:
image

I still couldn't figure out why I am getting this error:
image

Quillraven:
I noticed that the javadoc file was missing in the publication. I read somewhere that dokka does configure everything automatically in KMP but seems like this is not the case. Added some stuff to the build.gradle.kts file taken from a tutorial and now I see a javadoc file in the publication but seems like this is also not working properly because in a "normal" project, where I can add Fleks KMP as a dependency, I don't get any documentation in the IDE.

So at least with dokka something is not properly setup yet. Maybe also something else is not properly setup and that's why it doesn't work with KorGE. 

Unfortunately a proper documentation for gradle and KMP is very rare and hard to find :D

Quillraven:
Okay, I think this is more an Intellij/gradle/MPP bug and not a bug in the setup we have because:
https://youtrack.jetbrains.com/issue/KT-45979/MPP-Gradle-Sources-and-Javadoc-of-published-Android-klibs-artifacts-not-accessible-in-IDEA-although-theyre-present-in-the

https://youtrack.jetbrains.com/issue/IDEA-227215/Maven-local-repo-sources-not-imported-in-Idea-libraries-after-Gradle-6-update


What I see is that when I use the published mavenLocal version in a non-KorGE project then I can use the code just fine. Decompiling/Javadoc doesn't work automatically but when I go into a decompiled class "e.g. by CTRL+clicking on the World) then I can choose the "sources" which are correctly published and available in the mavenLocal folders.
Once I select the correct sources jar then JavaDoc and also looking at the source files works normally as you'd expect it.

Therefore, I am pretty sure that the publish process is correct. I need to try it, if it makes a difference when not using `-SNAPSHOT` at the end of the version.

----

What I still don't get is, why KorGE won't work. Maybe also related to the `-SNAPSHOT` version name. When I have time, I will try that.

jobe-m:
I checked out locally the "next" branch and did a local release "1.3.1-KMP-SNAPSHOT" of it.

Then I imported it in my Korge game - this was working. I tested jvm, linux x86 native and js. I even could compile it for Android but could not yet load it on the device.
Finally I changed the ecs-fleks example in Korge to use directly the "1.3.1-KMP-SNAPSHOT" version from maven local and that also worked for above mentioned targets.

BTW I am using Intellij IDEA version 2021.3 (Build #IC-213.5744.223, built on November 27, 2021) if this makes a difference...

Quillraven:
That's interesting :D I use 2022.1 but I will try a different one on the weekend. You did not change anything of the next branch and simply used `publishToMavenLocal` ?

Thanks for trying it out and cool to hear that at least it is working just fine on your end! I will make a `1.3-KMP-RC1` release then on the weekend to verify that the publishing to maven also works with the new setup. If that is fine, then let's release a normal release in the upcoming weeks - cool stuff :)

src/commonMain/kotlin/com/github/quillraven/fleks/system.kt
```diff
@@ -138,26 +130,25 @@ object Manual : SortingType
...
+    anyOfComponents: Array<KClass<*>>? = null,
     private val comparator: EntityComparator = EMPTY_COMPARATOR,
     private val sortingType: SortingType = Automatic,
     interval: Interval = EachFrame,
     enabled: Boolean = true
 ) : IntervalSystem(interval, enabled) {
     /**
      * Returns the [family][Family] of this system.
-     * This reference gets updated by the [SystemService] when the system gets created via reflection.
+     * This reference gets updated by the [SystemService] when the system gets created via the SystemFactory.
```
Quillraven, line 143: This comment is wrong

jobe-m:
> You did not change anything of the next branch and simply used `publishToMavenLocal` ?

Yes, I took the next branch unchanged and used `publishToMavenLocal`.

> I will make a `1.3-KMP-RC1` release then on the weekend to verify that the publishing to maven also works with the new setup. If that is fine, then let's release a normal release in the upcoming weeks - cool stuff :)

Having a release candidate in maven central would be nice also to test using it in Korge. Thanks!

Quillraven:
We seem to have a problem in the `addRemove` benchmark, since it  is a lot slower than the JVM version. From a quick look it seems there are a lot of `.simpleName` calls which are causing that. I will have a look in more detail tomorrow. The simple and complex benchmark are running fine as before with the same performance.

----

Also, github actions seem to fail to publish the artifact because of dokkaHtml. I can run the tasks locally on my computer without any issues, so I am not quite sure what is wrong.

I try to run it once more now with `windows-latest` instead of `ubuntu-latest`. Let's see!

Quillraven:
@jobe-m : I was finally able to publish it to mavenCentral. At least I got no more errors in the nexus repository :D
This means we should have a `1.4-KMP-RC1` version soon that we can test. I changed the version to 1.4 instead of 1.3 because of the additional DSL feature.

jobe-m:
Great, thanks. I will check the new RC1 version soon.

jobe-m:
I checked the `1.4-KMP-RC1` version in my game and also in a new plain "Hello World" project which is derived from the Korge "Hello World" repo and both are working. The "Fleks Hello World" example you can found here: https://github.com/jobe-m/korge-fleks-hello-world

PR #55, opened 2022-06-21
Small fix in "System" documentation

mariorez:
Sets in the description the same system type used in the code example.


Quillraven:
Thanks for the fix! It is indeed wrong ;)

I was always thinking of moving this documentation to the Wiki area instead of the ReadMe.md but I am not sure what people prefer.

What do you think?

Quillraven:
Then I will try to move that to the wiki in the upcoming days. By the way, good job with your games! Do you mind if I link them as well as reference games for Fleks?

Quillraven:
FYI I started with the wiki a little bit and your two games are mentioned: https://github.com/Quillraven/Fleks/wiki/Examples
@jobe-m : Shall I also add your KorGE example to that area? Or do you prefer a different example for a KorGE KMP demonstration?

Issue #59, opened 2022-07-11
Family numEntities returns a wrong value


Quillraven:
I just saw that `numEntities` of a `Family` is returning a wrong value. This method is not really used inside Fleks, that's why it is not a critical bug imo but still we need to fix it.

The reason is that it returns the length of the `BitArray` which is the index of the highest set bit. This is working for our test cases with just one Entity but it is wrong in other cases.
E.g. adding just one entity with ID 7 will return a size of 8 instead of 1.

```kotlin
 @Test
fun foo() {
    val bits = BitArray()
    bits.set(7)
    println(bits.length())  // prints 8
}
```

I see two possible solutions:
- add a `size` function to `BitArray` that returns the amount of set bits
- use the family's `entitiesBag` and return the `size` of that bag

The second solution is faster but has a certain delay until it represents the correct number. This happens inside `updateActiveEntities`:
```kotlin
internal fun updateActiveEntities() {
    if (isDirty) {
        isDirty = false
        entities.toIntBag(entitiesBag)
    }
}
```

The first solution always returns the correct number but can be quite expensive in case of a lot of entities. One implementation could look like this:

```kotlin
val numBits: Int
	get() {
		var sum = 0
		for (word in bits.size - 1 downTo 0) {
			val bitsAtWord = bits[word]
			if (bitsAtWord != 0L) {
				for (bit in 63 downTo 0) {
					if ((bitsAtWord and (1L shl (bit % 64))) != 0L) {
						sum++
					}
				}
			}
		}
		return sum
	}
```

I personally prefer the first option because I think `numEntities` will not get called by users a lot per frame and it returns the expected result immediately all the time.

@jobe-m : What do you think?

jobe-m:
I would also vote for the first solution and mark the function that it is expensive in the documentation.

Quillraven:
part of 1.5 versions that are coming soon

PR #60, opened 2022-07-17
(#59) fix family numEntities

Quillraven:
- fix numEntities
- add isEmpty and isNotEmpty
- add first and firstOrNull

----

I also added a few utility functions to a family to easily check if it is empty or not. Also, it can be useful to get just a random single entity out of a family and do something with it. For that I introduced now `first` and `firstOrNull`.
My use case was that I check in a system if the family is not empty then I want to do something with the first entity. I did that e.g. for a camera system that updates the game camera and optionally can lock the camera to one specific entity.

Before I had to do an iteration via `forEach`. Now this can be solved via a call to `first`.

I will make the same changes to the KMP branch in the upcoming days and release a 1.4.1 version for both JVM and KMP.

### DIFF
--- a/src/main/kotlin/com/github/quillraven/fleks/collection/bag.kt
+++ b/src/main/kotlin/com/github/quillraven/fleks/collection/bag.kt
@@ -86,8 +86,27 @@ class IntBag(
     val capacity: Int
         get() = values.size
 
+    private val isEmpty: Boolean
+        get() = size == 0
+
     val isNotEmpty: Boolean
         get() = size > 0
+
+    val first: Int
+        get() {
+            if (isEmpty) {
+                throw NoSuchElementException("Bag is empty!")
+            }
+            return values[0]
+        }
+
+    val firstOrNull: Int?
+        get() {
+            if (isEmpty) {
+                return null
+            }
+            return values[0]
+        }
 
     fun add(value: Int) {
         if (size == values.size) {
--- a/src/main/kotlin/com/github/quillraven/fleks/collection/bag.kt
+++ b/src/main/kotlin/com/github/quillraven/fleks/collection/bag.kt
@@ -86,7 +86,7 @@ class IntBag(
     val capacity: Int
         get() = values.size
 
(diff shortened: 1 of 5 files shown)


jobe-m (approved):
Looks good to me :)

Quillraven:
I need to check the implementation of first in more detail because I think there might be issues when you iterate through a family with forEach and calling first on that same family while iterating.

I wasn't able to break it yesterday but I have some ideas now on how to break it.

I am not sure if we need to consider that because I cannot imagine a scenario where that makes sense but let me think about it some more ;)

jobe-m:
No problem. Take your time.
I am in the meantime now also using FamilyListeners to initialize entities with a specific configuration of components. I like that approach very much. :)

Quillraven:
I think I found a good solution and yes, indeed I was able to break it. 
Removal of entities during iteration was actually not a problem because it is happening delayed. 
Adding is also sometimes fine because according to the generated bytecode that I saw, Kotlin is getting the size of an array before it starts looping when you write `array.forEach` which is nice because it then uses the size before any modification during iteration.

With a combination of both I was able to break it as seen in the `WorldTest` test case but I changed the implementation of `first` and `firstOrNull` to fix that. We can use the `delayRemoval` flag of the entity service to check if an iteration is currently happening. In those cases we simply don't update the underlying entitiesBag and assume it is up to date because it an iteration is already in process.

There could be a potential second special case where users are iterating over a family and call `first` on another family. If the entities of that second family is still empty then it is also save to call `updateActiveEntities`.

The "tricky" part is that we actually need to call `updateActiveEntities` because it is possible to only make calls to `first` or `firstOrNull` and we need to update the family in those cases because otherwise those calls will never return anything. Until now this was only needed in the `forEach` function of the family.

Hope this makes sense :D I will leave this PR open for another day or two. If we don't see any issues with that then I will merge it and prepare the new version.

Quillraven:
part of 1.5-JVM

PR #61, opened 2022-07-22
(#59) fix family numEntities (KMP)

Quillraven:
- add isEmpty and isNotEmpty
- add first and firstOrNull

----

same changes as in JVM


jobe-m (approved):
Looks good to me!

Quillraven:
will release 1.5-KMP when the JVM version is released on maven because I don't know how maven behaves when I publish two releases of the same artifact :D To be safe I want to wait.

PR #67, opened 2022-08-17
Kmp 1.6 changes

Quillraven:
1.6 changes for KMP


jobe-m:
Thanks for the release! I will try the new snapshot functionality. Sorry, I did not realize the review request earlier. It looks good to me 👍

Quillraven:
> Thanks for the release! I will try the new snapshot functionality. Sorry, I did not realize the review request earlier. It looks good to me 👍

No worries ☺️ There shouldn't be any major issues and in case we find something. We can always release a 1.6.1 version.

I just wanted to have both JVM and KMP on the same version.

PR #68, opened 2022-08-17
break through new Component/ComponentType approach

Quillraven:
This is just a draft PR for potential Fleks 2.0 changes.

At the moment it is a quick breakthrough implementation of a new `Component` and `ComponentType`. That way we can remove the reflection stuff for components completely for JVM. For KMP it is no longer necessary to register a factory method for components.

Discussion continues in the "Discussions" area for Fleks ;)

example of new API:

```kotlin
// new Component interface; requires to override the type() function to return a ComponentType
data class Position(var x: Float = 0f, var y: Float = 0f) : Component<Position> {
   companion object : ComponentType<Position>()

   override fun type() = Position
}

// we can keep existing stuff of course like ComponentListener. They will still work as before
class PositionComponentListener : ComponentListener<Position> {
    override fun onComponentAdded(entity: Entity, component: Position) {
        println("added")
    }

    override fun onComponentRemoved(entity: Entity, component: Position) = Unit
}

// No need to inject ComponentMapper anymore; we can directly access components of an entity in a system like
// in this example via "entity[Position]"
@AllOf([Position::class])
class PositionSystem : IteratingSystem() {
        println(entity[Position])
    }
}

fun main() {
    val w = world {
        components {
            add<PositionComponentListener>()
        }

        systems {
            add<PositionSystem>()
        }
    }

    w.entity {
        // components are no longer created via reflection. They are directly created and added by the user
        it += Position(2f, 3f)
    }

    w.update(1f)
}
```

### DIFF
--- a/src/main/kotlin/com/github/quillraven/fleks/component.kt
+++ b/src/main/kotlin/com/github/quillraven/fleks/component.kt
@@ -6,6 +6,18 @@ import java.lang.reflect.Constructor
 import kotlin.math.max
 import kotlin.reflect.KClass
 
+open class ComponentType<C>(val id: Int = nextId++) {
+    companion object {
+        private var nextId = 0
+    }
+}
+
+interface Component<C> {
+    fun type(): ComponentType<C>
+
+    fun onRemove(entity: Entity) = Unit
+}
+
 /**
  * Interface of a component listener that gets notified when a component of a specific type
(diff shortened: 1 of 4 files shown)


Quillraven:
I will close this one and open a new one that is based on the KMP branch

PR #69, opened 2022-09-14
Fleks 2.0

Quillraven:
Progress:
- [x] Breakthrough
- [x] Code adjustments
- [x] Unit test adjustments
- [x] Update comments
- [x] Update wiki
- [x] Update ReadMe
- [x] Update version and combine JVM and KMP. Maybe make the current master a new JVM_history branch to keep the history  
- [x] Verify new version in Dinoleon and Mystic Woods

----

This is currently a WIP for Fleks 2.0. I did most of the code adjustments  but I am still missing the unit test adjustments and comments.

I did a test driven development approach (`fleks2-tdd.kt`) which I kinda liked for this exercise. It helped to get the API faster in the way I wanted it.

Also, I want to discuss the changes that I have applied, if they make sense to everyone or not. Be aware, here comes a wall of text ;) Questions are highlighted in bold and with a ❓ emoji.

To easily create the boilerplate code for the new Fleks component approach, I suggest to create a Live Template in the IDE. For example in IntelliJ we can do the following. You then just type `flekscomponent` or whatever abbreviation you want to use to get the necessary code.
image

<details><summary><h4>Template text</h4></summary>
<p>

```Kotlin
class $FLEKS_COMPONENT$ : Component<$FLEKS_COMPONENT$> {
	override fun type() = $FLEKS_COMPONENT$

	companion object : ComponentType<$FLEKS_COMPONENT$>()
}
```

</p>
</details>

**TLDR**: imo the Fleks codebase gets a little smaller and easier to understand with those changes. Also, with the removal of reflection and factory methods it is easier to understand, use and debug. The performance is slightly faster than before. I guess because of the removal of factory method calls and the conversion from listeners to hooks.

----

# Changes

## Components

It is no longer necessary to add components to the `WorldConfiguration`. `ComponentMapper` are created automatically similar to the JVM version, and we also no longer need any reflection or factory methods.

In order for that to work, a new interface and abstract class was added and every component now must implement the new `Component` interface. 
More about the benefits below in the "Systems" section.

A position component now looks like this:

```Kotlin
data class Position(
    var x: Float,
    var y: Float,
) : Component<Position> {
    override fun type(): ComponentType<Position> = Position

    companion object : ComponentType<Position>()
}
```

The `ComponentType` is important to give each type of component a unique ID which is necessary for Fleks internal stuff to optimize the way components are stored.

With Kotlin's unnamed companion object we can then achieve imo some nice syntax sugar (see below in "Systems" section).

With this change it is now also possible to add the same component multiple times to an entity. This is not possible in Fleks 1.x. **This is something new!** :) Here is an example of a Sprite component that can either be a background or foreground of an entity:

```Kotlin
ata class Sprite(
    val background: Boolean,
    var path: String = "",
) : Component<Sprite> {
    override fun type(): ComponentType<Sprite> {
        return when (background) {
            true -> SpriteBackground
            else -> SpriteForeground
        }
    }

    companion object {
        val SpriteForeground = object : ComponentType<Sprite>() {}
        val SpriteBackground = object : ComponentType<Sprite>() {}
    }
}
```

Because `type()` is a function, we can basically do whatever we want and therefore reuse the same component class for different purposes.

## Systems

Since we cannot get annotations in KMP, I adapted the existing approach of defining the family configuration in a system's constructor.
For that, I added a new `family` function with a new `FamilyDefinition` configuration to define the family and removed the allOf, anyOf, noneOf arguments with a single family argument.

This has two advantages imo:
1. Users must provide a family and cannot forget it anymore. Before this was not immediately visible because allOf, anyOf and noneOf had a default to an empty collection.
2. The amount of constructor arguments for an `IteratingSystem` gets reduced

Here is an example of the new syntax:

```Kotlin
private class PositionSystem : IteratingSystem(family { all(Position) }) {
        entity[Position].x++ // <-- new way of retrieving components
    }
}
```

Note, that `ComponentMapper` are no longer necessary from a user's point of view. Via `entity[Position]` we can now simply access the position component of an entity.

Also, family definitions no longer need the `::class` syntax. The `allI`, `none` and `any` functions of a `FamilyDefinition` take a vararg of `ComponentType`.

❓ **Shall we use the familyDefinition approach? Or is the previous approach nicer?**

----

The `configureEntity` function got replaced with an extension function for an `Entity`. Instead of writing `configureEntity(entity) { ... }` it is now `entity.configure { ... }`. This is a minor change but I think it is easier to read. Also, the configuration API got changed from mappers to the new component approach.
Here is an example of one of the benchmark systems that removes a FleksPosition component and adds a FleksLife component:

```Kotlin
class FleksSystemComplex1 : IteratingSystem(family { all(FleksPosition).none(FleksLife).any(FleksSprite) }) {
    private var actionCalls = 0

        if (actionCalls % 2 == 0) {
            entity[FleksPosition].x++
            entity.configure { it += FleksLife() } // <-- new way of adding components
        } else {
            entity.configure { it -= FleksPosition } // <-- new way of removing  components
        }
        entity[FleksSprite].animationTime++
        ++actionCalls
    }
}
```

Finally, a new `compareEntityBy` function was added for easier sorting. You can now make any component implement the `Comparable` interface and then write the sorting code in that specific component. With that you can achieve something like this:

```Kotlin
private data class SystemTestComponent(
    var x: Float = 0f
) : Component<SystemTestComponent>, Comparable<SystemTestComponent> {
    override fun type() = SystemTestComponent

    override fun compareTo(other: SystemTestComponent): Int {
        return other.x.compareTo(x)
    }

    companion object : ComponentType<SystemTestComponent>()
}

private class SystemTestIteratingSystemSortManual : IteratingSystem(
    family = family { all(SystemTestComponent) },
    comparator = compareEntityBy(SystemTestComponent), // <-- easy and concise way to create an EntityComparator
    sortingType = Manual
) {
    // ... rest of the code omitted
}
```

## Entities

As seen above, entity configuration and also the creation got adjusted to the new components implementation. Here are some examples:

```Kotlin
    @Test
    fun testComponentHooks() {
        val addComponent = Position(0f, 0f)
        val removeComponent = Position(0f, 0f)
        lateinit var testWorld: World
        testWorld = world {
            components {
                // example of an onAdd hook for a position component. 'component' is of type Position
                onAdd(Position) { world, entity, component ->
                    component.x = 1f
                    assertEquals(testWorld, world)
                    assertTrue { entity.id in 0..1 }
                    assertTrue { component in listOf(addComponent, removeComponent) }
                }

                // example of an onRemove hook for a position component. 'component' is of type Position
                onRemove(Position) { world, entity, component ->
                    component.x = 2f
                    assertEquals(testWorld, world)
                    assertEquals(Entity(1), entity)
                    assertEquals(removeComponent, component)
                }
            }
        }

        // entity that triggers onAdd hook
        testWorld.entity { it += addComponent }
        // entity that triggers onRemove hook
        val removeEntity = testWorld.entity { it += removeComponent }
        testWorld.configure(removeEntity) { it -= Position }

        assertEquals(1f, addComponent.x)
        assertEquals(2f, removeComponent.x)
    }

    @Test
    fun testFamilyHooks() {
        lateinit var testWorld: World
        lateinit var testFamily: Family
        var numAddCalls = 0
        var numRemoveCalls = 0
        testWorld = world {
            testFamily = family { all(Position) }
            families {
                // example of an onAdd hook for a family with allOf=Position
                onAdd(testFamily) { world, entity ->
                    ++numAddCalls
                    assertEquals(testWorld, world)
                    assertTrue { entity.id in 0..1 }
                }

                // example of an onRemove hook for a family with allOf=Position
                onRemove(testFamily) { world, entity ->
                    ++numRemoveCalls
                    assertEquals(testWorld, world)
                    assertEquals(Entity(1), entity)
                }
            }
        }

        // entity that triggers onAdd hook
        testWorld.entity { it += Position(0f, 0f) }
        // entity that triggers onRemove hook
        val removeEntity = testWorld.entity { it += Position(0f, 0f) }
        testWorld.configure(removeEntity) { it -= Position }
        // trigger family update to call the hooks
        testFamily.updateActiveEntities()

        assertEquals(2, numAddCalls)
        assertEquals(1, numRemoveCalls)
    }
```

## ComponentListeners and FamilyListeners

I thought about this and in my opinion it is actually useless to have a list of listeners because why would someone register multiple listeners for the same component/family? At least I never used that and I could not come up with why I would ever do that.

That's why I changed the listener approach to a "hook" approach. It is basically the same but you specify only one add/remove hook per component/family. That way we don`t need to iterate over a list of listeners all the time which always has the size 1.

**Important to note** is that a hook always gets the world instance as an argument and the world now keeps the registered injectables and does not remove them anymore after creation. That way you can still get any relevant object into your hook code.

Here is an example of a component and a family hook. Fleks distinguishes between an `add` and `remove` hook. It is not necessary to add the code directly to the WorldConfiguration. It can of course link to a function in a different file.

```Kotlin
    @Test
    fun testComponentHooks() {
        val addComponent = Position(0f, 0f)
        val removeComponent = Position(0f, 0f)
        lateinit var testWorld: World
        testWorld = world {
            components {
                // example of an onAdd hook for a position component. 'component' is of type Position
                onAdd(Position) { world, entity, component ->
                    component.x = 1f
                    assertEquals(testWorld, world)
                    assertTrue { entity.id in 0..1 }
                    assertTrue { component in listOf(addComponent, removeComponent) }
                }

                // example of an onRemove hook for a position component. 'component' is of type Position
                onRemove(Position) { world, entity, component ->
                    component.x = 2f
                    assertEquals(testWorld, world)
                    assertEquals(Entity(1), entity)
                    assertEquals(removeComponent, component)
                }
            }
        }

        // entity that triggers onAdd hook
        testWorld.entity { it += addComponent }
        // entity that triggers onRemove hook
        val removeEntity = testWorld.entity { it += removeComponent }
        testWorld.configure(removeEntity) { it -= Position }

        assertEquals(1f, addComponent.x)
        assertEquals(2f, removeComponent.x)
    }

    @Test
    fun testFamilyHooks() {
        val testFamilyDef = familyDefinition { allOf(Position) }
        var numAddCalls = 0
        var numRemoveCalls = 0
        lateinit var testWorld: World
        testWorld = world {
            families {
                // example of an onAdd hook for a family with allOf=Position
                onAdd(testFamilyDef) { world, entity ->
                    ++numAddCalls
                    assertEquals(testWorld, world)
                    assertTrue { entity.id in 0..1 }
                }

                // example of an onRemove hook for a family with allOf=Position
                onRemove(testFamilyDef) { world, entity ->
                    ++numRemoveCalls
                    assertEquals(testWorld, world)
                    assertEquals(Entity(1), entity)
                }
            }
        }

        // entity that triggers onAdd hook
        testWorld.entity { it += Position(0f, 0f) }
        // entity that triggers onRemove hook
        val removeEntity = testWorld.entity { it += Position(0f, 0f) }
        testWorld.configure(removeEntity) { it -= Position }
        // trigger family update to call the hooks
        testWorld.family(testFamilyDef).updateActiveEntities()

        assertEquals(2, numAddCalls)
        assertEquals(1, numRemoveCalls)
    }
```

❓ **Shall we keep the hook approach? Or shall we go back to the listeners approach?**

## Injectables

As mentioned before, injectables are no longer cleared after the world is created. The main reason is the new hook approach.

But there are also other changes like e.g. the removal of the `Inject` object. Since the world is now keeping references to the different injectables, we can directly use the world to inject stuff. 
Also, it is now possible to do constructor DI instead of just property DI. Imo constructor DI is always preferable because it creates cleaner code and makes it easier to test classes when you directly see their dependencies. 
In order for that to work, the world creation process was adjusted a little bit and a new `inject` function was added.

Here is an example of a system that showcases both DI mechanisms:

```Kotlin
private class SpriteSystem(
    val cstrInjectable: String = inject() // <-- constructor DI via the global inject function
) : IteratingSystem(family { any(SpriteBackground, SpriteForeground) }) {
    val propInjectable: String = world.inject("qualifiedString") // property DI via world.inject. The global function will work here as well

    override fun onTickEntity(entity: Entity) = Unit
}
```

## WorldConfiguration

As already seen in some examples above, the configuration of the world has also changed slightly because of the removal of the listeners **but** also, because of the removal of factory methods for systems. Systems are now directly created which makes it imo easier to understand the code and also to debug the creation process, if necessary.

It is important, that the `systems` block comes last in the configuration when family or component hooks are registered. Fleks takes care of that and throws a new `FleksWrongConfigurationOrder` exception if users fail to do so.
The reason is that system constructors can create new entities and if user hooks are registered after system creation, then of course they are not called correctly for such entities.

Here is an example test case with component hooks and a system:

```Kotlin
    @Test
    fun notifyComponentHooksDuringSystemCreation() {
        var numAddCalls = 0
        var numRemoveCalls = 0

        world {
            components {
                onAdd(WorldTestComponent) { _, _, _ -> ++numAddCalls }
                onRemove(WorldTestComponent) { _, _, _ -> ++numRemoveCalls }
            }

            systems {
                add(WorldTestInitSystem())
            }
        }

        assertEquals(1, numAddCalls)
        assertEquals(0, numRemoveCalls)
    }
```


jobe-m:
I just read through the text. It sounds awesome. I think I will like the new API very much :)
Is it possible to build this branch and release locally and use it already at this state? I would like to get a feeling for the changes by adapting my code where I implemented an animation script system with fleks.

Quillraven:
Thank you for checking it out and for your feedback!

I think it should work but comments are not updated yet and I want to bring in the old mapper API for update entities to make it easier for people to move from 1.x to 2.x. Then at least inside system code you don't have to migrate immediately to the new API.

I will try to do some tests on my own next week. I let you know if it worked or not and if it does work I will convert this PR draft to a real PR 😉

jobe-m:
To be honest I wouldn't need the old mapper API and I would not miss it :)
I was always a bit confused when I wanted to change some component in an existing entity.
First one need to call `configureEntity` and than "search" for the correct mapper and call `addOrUpdate` on it.

Does the new `entity.addOrUpdate {...}` API already work? It looks so much more streamlined to the actual process what one want to do :)

> ❓ Shall we use the familyDefinition approach? Or is the previous approach nicer?

The familyDefintion approach looks nicer IMO.

> ❓ Shall we keep the hook approach? Or shall we go back to the listeners approach?

Definitely go with the new hook approach. It looks more intuitive to me!

Quillraven:
Indeed everything should work already since the tests are passing. 

About the mapper API: I personally prefer the new API with ComponentType and imo it is fast enough from performance point of view. Therefore, I'd remove the mapper API completely to "limit" users with options and don't overwhelm them.

However, there might be some bigger projects already existing and migrating to the new API could be time consuming. So maybe we should keep it still for 2.0 but mark it with a deprecation warning and remove it completely in the next release?

jobe-m:
Ok, sounds good. Indeed good idea to give a deprecate warning :)

jobe-m:
Currently in an `onAdd` hook of a family I cannot use `entity["Component"]` as the "get" operator for Entity is only implemented in an IteratingSystem. Will this be somehow also possible in the hooks or do we need here the "old" mapper API to get a component from an entity?

Quillraven:
I think world.configure(entity) should do the trick but you have a point. Maybe hooks should already run within an EntityUpdateConfiguration scope to be able to easily modify the entity.

I will have a look at that 🙂

jobe-m:
No, it doesn't work. Because there is no "Entity.get" implemented in EntityUpdateCfg class.

image

Quillraven:
@jobe-m: Please check out the last commits. It is now possible to easily access components via the get operator inside a component/family hook, entity create block or entity update block.

I also fixed the WorldCfg DSL so that nesting of `components`, `families` and `systems` blocks is no longer possible. The compiler will now throw an error if a user does that. I used the @DslMarker annotation wrongly until now.

I hope this is now more intuitive!

jobe-m:
Sound good. I will check it out :)

jobe-m:
Ah, I didn't do a reload of the pull request page. Now I see your edits.
Great. I like in "edit2" the onAddPosition syntax. Even better with the type alias. :)

jobe-m:
What do you think if the component hooks are added to the component data class? 
I guess that way the hook variables do not pollute the global name space. And the code would be visually located where the data is.
You could even expand the Intellij Template. Thus the code is already there and if not needed can be easily deleted again or not added to the components section in the world.

Here an example how it looks in one of my components:
```kotlin
/**
 * This component holds all needed details to animate an entity's component.
 */
data class AnimationScript(
    var debugAnnotate: Boolean = false,  // TODO not implemented yet
    var tweens: List<TweenBase> = listOf(),

    // Internal runtime data
    var index: Int = 0,            // This points to the animation step which is currently in progress
    var timeProgress: Float = 0f,  // Elapsed time for the object to be animated
    var waitTime: Float = 0f,
    var active: Boolean = false
) : Component<AnimationScript> {
    override fun type(): ComponentType<AnimationScript> = AnimationScript

    companion object : ComponentType<AnimationScript>() {
        /**
         * Initialize internal waitTime property with delay value of first tweens if available.
         */
        val onComponentAdded: ComponentHook<AnimationScript> = { world, entity, component ->
            if (component.tweens.isNotEmpty()) component.waitTime = component.tweens.first().delay ?: 0f
        }

        val onComponentRemoved: ComponentHook<AnimationScript> = { world, entity, component ->
            // do something
        }
    }
}
```

In the world it could then look like this:

```kotlin
import components.*
import components.AnimationScript.Companion.onComponentAdded
import components.AnimationScript.Companion.onComponentRemoved

[...]

        components {
            onAdd(AnimationScript, onComponentAdded)
            onRemove(AnimationScript, onComponentRemoved)
        }

```

Quillraven:
I think this is totally valid and fine but it is up to the users where they want to put the code. I personally would follow the same approach as you because of personal preference ;)

jobe-m:
My last commit makes the second line possible within an EntityHookContext. I think this is more readable :)

image

jobe-m:
I will do the same for `getOrNull()`. The first looks much better :)

image

src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
```diff
@@ -20,6 +20,10 @@ class EntityHookContext(
     inline operator fun <reified T : Component<*>> Entity.get(type: ComponentType<T>): T {
         return compService.holder(type)[this]
     }
+
+    inline fun <reified T : Component<*>> Entity.contains(componentType: ComponentType<T>): Boolean {
```
Quillraven: I suggest to make it an operator fun then you can also use the `in` syntax if you want.

src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
```diff
@@ -21,8 +21,12 @@ class EntityHookContext(
         return compService.holder(type)[this]
     }
 
-    inline fun <reified T : Component<*>> Entity.contains(componentType: ComponentType<T>): Boolean {
-        return compService.holder(componentType).contains(this)
+    inline fun <reified T : Component<*>> Entity.contains(type: ComponentType<T>): Boolean {
+        return compService.holder(type).contains(this)
+    }
+
+    inline fun <reified T : Component<*>> Entity.getOrNull(type: ComponentType<T>): T? {
```
Quillraven, line 28: `contains` and `getOrNull` sound reasonable. We also need to add them to `IteratingSystem` and `Family`. I think we could avoid duplicating those methods by making the system and family also a `EntityHookContext` and delegate it to the world's hook instance but as far as I know this simply implements the methods in those classes by then calling the delegation instance (=hook instance of the world). So it is one additional function call and that's why I hesitate to use that approach.

But duplicating those "basic" entity extension functions in different places is also somehow annoying :D

src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
jobe-m: So that means removing the `inline`ing? Didn't know that. I need to check the `in` syntax in my code...

src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
jobe-m: Yes, I also saw that both functions would be also useful in the system code. I already added them locally but then hadn't time to go on ;)
Yes, it looks annoying to have these functions spread around the codebase but they are luckily small and I also have no better idea :)

src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
jobe-m: Ah now I got it ;)

Quillraven:
Added `Entity.get`, `Entity.getOrNull` and `Entity.contains` now in the contexts, family and system and I actually identified a wrong behavior for contains calls inside a component removehook. It returned true although it should be false of course because the components gets removed ;)

jobe-m:
>    if(entity has Sprite) {}

Very good idea. I like it, too.

> 2. [...]
>    I added a new `query` function to the world which I use already as a replacement in some unit test code. It executes a query to the world within an `EntityHookContext` for a given entity and returns the result of the query.

The new query function looks interesting. Is the "query" more expensive than if the same functionality is used inside a system or the other scopes? (I mean more expensive in terms of runtime e.g.)

I have quite some code which creates and configures entities outside of the world or the systems. I want to keep the entity creation outside of systems since entities are depending on e.g. the game level and I do not want to have a dependency of my systems to specific parts (levels) of my game. (At least I try so...)

Also I am adding a component to an entity in a familyHook function. (That added component is not part of the family definition for the FamilyHook to not trigger recursive calls of the hook).
Also in that FamilyHook I am using injected objects. 

I have already seen that this makes a problem with using injected objects in hook context. I get there an exception that an injected object is not used, because I do not use it on creation time of the systems (?). So I guess that this comes from the fact that the inject() is used later at runtime when an entity is created.

Thus currently I need to comment out the `throw FleksUnusedInjectablesException(...)` in world.kt to get my code working...

Maybe we need another mechanism to use "world external" objects inside the hook functions?

>    // note that we are outside of a family, system or hook scope and only have a world
>    // 1) We can get the component of an entity
>    val pos: Position = world.query(entity) { it[Position] }
>    // 2) We can also check if an entity has a component
>    if( world.query(entity) { it has Position }) {

This again looks nice. I will try it out soon...

> 3. Lastly, the `EntityComparator` functionality had to change a little bit. Both `compareEntity` and `compareEntityBy` take a world argument now which defaults to the `CURRENT_WORLD` which means a user does not need to provide anything when using those functions within a WorldConfiguration block. `compareEntity` now then also simply runs within an `EntityHookContext` to provide the base entity extension functions to easily retrieve components.

Currently I have no use case yet for sorting of entities in my systems. So I am not using it and have no experience yet.

Quillraven:
I think it is basically the same as Kotlin's `let`, `apply`, `run`, etc.. I don't see any expensive overhead in the `query` functionality. Should be the same as within a system.

You have a good point again with the injectables. I did not think about that yet. Shall we simply remove this check completely and provide a function in the world like `getUnusedDependencies`? Then users can decide if they want to analyze and debug that or not.
But for sure it is no longer an error because as you mentioned this does not make sense with hooks.

jobe-m:
> I think it is basically the same as Kotlin's `let`, `apply`, `run`, etc.. I don't see any expensive overhead in the `query` functionality. Should be the same as within a system.

👍 

> You have a good point again with the injectables. I did not think about that yet. Shall we simply remove this check completely and provide a function in the world like `getUnusedDependencies`? Then users can decide if they want to analyze and debug that or not. But for sure it is no longer an error because as you mentioned this does not make sense with hooks.

Good idea to have that checking function. I had the idea to transform the exception into a warning message. But I guess that it does not need a warning since the inject functionality is now different. The injected objects are there forever and can be used whenever needed. :)

Quillraven:
I updated Dinoleon to 2.0 and everything works out! FPS are also the same although I did not optimize some things with the migration (runs at stable ~1300 FPS on my machine before 2.0 and after).

I did not push the changes yet because I want to wait for the official 2.0 maven release but so far looks good!

Will try to update MysticWoods tomorrow and if that also looks good then I will update the wiki (fun! :P).

Maybe we can already release 2.0 in 1 or 2 weeks.

jobe-m:
Awesome 👍😎

I did not have luck in updating my code to 2.0. It compiles successfully, but when I run it my animation system does nothing. It looks I did some mistake while adapting the code. I need still to debug it ;)

Quillraven:
MysticWoods is also working and runs ~50FPS faster than before (from 1170FPS to 1220FPS).

What I like is that some `lateinit var` in components can be changed to `val` since a component is no  longer created via reflection and therefore I can immediately set certain values. Same for some `var` values that can also be changed to `val`.

----

I also need to use the `query` function a few times in this project because I have an entity AI wrapper class for AI stuff. It wraps a world and an entity and provides some utility functionality for AI. Since mappers are gone but I still need to check for some components or access them, I need to use the `query` function. I ended up writing my own `Entity.` extension functions using the `world.query` implementation internally inside that AIEntity class.

I also had to use that in a `PlayerController` which is responsible for player input handling and a `GameModel` that I use for the UI in an MVVM approach.

----

What I did not like that I cannot negate the new infix `has` function. From what I read this is simply not supported by Kotlin (yet?). I was surprised since you can already write `!in` or `!is` in standard Kotlin but `!has` did not work for me. So, I ended up writing it like this `if (!(entity has AnimationComponent)) `.

----

@jobe-m: let's wait until you also were able to migrate your application to 2.0 to be sure that there is no problem in the code changes.

@czyzby: I have two questions for you again where you might be able to help please. First of all, thank you for the different suggestions you proposed early in the development of Fleks. With the new 2.0 version I used some of them :)

1. **Is it possible to support negation of infix functions that return a boolean?** For example I have a `has` function for an entity to return true if it has a specific component or not. I can write `entity has MoveComponent`  but I cannot write `entity !has MoveComponent`. 
Is this somehow possible? Or can we only write `!(entity has MoveComponent)`?
3. Within the context of Fleks (like systems, families or hooks) it is possible to provide entity extension functions e.g. to add a component or remove a component. We can then write `entity += MoveComponent(speed =3f)` or `entity -= MoveComponent`. This is possible because systems, families and hooks internally have a reference to the component service and can therefore interact with it to get component information.
However, sometimes you write code outside of Fleks like e.g. some special event handling of your game. I did not see an option to provide those entity extension functions outside because an entity is just an ID and therefore has no access to anything.
What I ended up with is defining a `query` function to the fleks `World` which those external classes/logics should have access to. The `query` then allows the support of the entity extensions again. Here is an example of my ViewModel that reacts on a so called "AggroEvent":
    ```kotlin
    // ...
    is EntityAggroEvent -> {
                val source = event.aiEntity
                val sourceType = world.query(source) { it.getOrNull(AnimationComponent) }?.atlasKey
                val target = event.target
                val isTargetPlayer = world.query(target) { it has PlayerComponent }
                if (isTargetPlayer && sourceType != null) {
                    updateEnemy(source)
                }
            }
    ```

    And here is how I defined the `query` function of the world in Fleks. The `EntityHookContext` is the thing that provides the entity extension functionality:

    ```kotlin
    /**
     * Executes and returns the result of a [query] for the given [entity].
     * This can be used to retrieve [components][Component] of the [entity] or check
     * if it has certain [components][Component].
     */
    inline fun <reified T> query(entity: Entity, query: EntityHookContext.(Entity) -> T): T {
        return query(hookCtx, entity)
    }
    ```

    **Do you see a better approach to solve this?**

czyzby:
@Quillraven 

> Is it possible to support negation of infix functions that return a boolean?

I don't think so. Here's the associated issue: [KT-5351](https://youtrack.jetbrains.com/issue/KT-5351). It's not ideal, but you might want to provide `hasNot` infix method instead.

> Do you see a better approach to solve this?

What about recommending the built-in `with`?

```kotlin
with(world) { entity has Component }
```

It does change `this` within the context, but I think it introduces less mental overhead when reading the code - you can see that something is being done with a specific entity in the context of the chosen world.

jobe-m:
With this change locally in component.kt file my game how also works :)

image

Quillraven:
@jobe-m: good to hear that your game is also working, nice! And I have nothing against changing the logic. I will do that over the weekend and adjust code, comments and tests.

@czyzby: thank you for your quick response and information. I will try out the `with(world)` version which has the downside that I need to copy&paste the extension functions to the world as well but it should work and make it easier to read, as you mentioned.

Or do you know a solution how to solve the following:
- We have some base entity extension functions which are available in certain contexts like systems, families, hooks and soon also world.
- Those functions are `Entity.get`, `Entity.getOrNull`, `Entity.contains` and `Entity.has`
    ```kotlin
    /**
     * Returns a [component][Component] of the given [type] for the [entity][Entity].
     *
     * @throws [FleksNoSuchEntityComponentException] if the [entity][Entity] does not have such a component.
     */
    inline operator fun <reified T : Component<*>> Entity.get(type: ComponentType<T>): T =
        compService.holder(type)[this]

    /**
     * Returns a [component][Component] of the given [type] for the [entity][Entity]
     * or null if the [entity][Entity] does not have such a [component][Component].
     */
    inline fun <reified T : Component<*>> Entity.getOrNull(type: ComponentType<T>): T? =
        compService.holder(type).getOrNull(this)

    /**
     * Returns true if and only if the [entity][Entity] has a [component][Component] of the given [type].
     */
    inline operator fun <reified T : Component<*>> Entity.contains(type: ComponentType<T>): Boolean =
        compService.holder(type).contains(this)

    /**
     * Returns true if and only if the [entity][Entity] has a [component][Component] of the given [type].
     */
    inline infix fun <reified T : Component<*>> Entity.has(type: ComponentType<T>): Boolean =
        compService.holder(type).contains(this)
    ```
- I cannot put them on a global scope, I guess, because they require the `ComponentService`.
- My idea then was to use Kotlin's delegation and create an interface with those four functions, one implementation and then in any class that provides those extensions, I simply let it extend the interface and delegate it to the implementation.
- However, this does not seem to work because I use the `reified` keyword and virtual functions cannot have this keyword, for whatever reason.
- Do you see any solution to solve that? Or do I need to stick to the current approach and copy&paste those functions everywhere, where needed?

czyzby:
@Quillraven From what I understand, the problem is that each class provides the `ComponentService` in a different way, but you cannot define an interface with an inline method that has a custom receiver. That's true, but the same limitation does not apply to abstract classes, if they're not a deal breaker. Here's a proof of concept:

```kotlin
typealias Entity = Int

// Example abstract class:
abstract class ComponentProvider {
  abstract val componentService: ComponentService

  // Example reified extension methods:
  inline fun <reified T> Entity.getOrNull() = componentService[this]
}

// Example implementations:
class ComponentService : ComponentProvider() {
  operator fun get(entity: Entity) = entity

  override val componentService: ComponentService
    get() = this
}

class World : ComponentProvider() {
  override val componentService = ComponentService()
}

// Usage:
fun example() {
  val entity: Entity = 0
  with(ComponentService()) { entity.getOrNull<String>() }
  with(World()) { entity.getOrNull<String>() }
  // Does not compile:
  // entity.getOrNull<String>()
}
```

You'd probably have to put this somewhere in your class hierarchy for every relevant class, but it beats code duplication.

Quillraven:
That works for family, world and hooks but unfortunately won't work for IteratingSystem as it already inherits from IntervalSystem and therefore cannot inherit from a second class. I will think about it some more over the weekend. Maybe I can restructure a little bit and otherwise, I can at least avoid code duplication in most areas.

Thanks!

src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
```diff
@@ -11,58 +11,59 @@ import com.github.quillraven.fleks.collection.compareEntity
...
- *
- *
- * Also, the same functionality is provided by a [Family] and an [IteratingSystem] but not via this class.
- * The methods must be manually copied in all those areas because unfortunately, I have no better idea how to solve that :(
+ * A class for basic [Entity] extension functions within an add/remove hook of a [Component], [Family],
+ * [IntervalSystem], [World] or [compareEntity].
  */
-@EntityCtxMarker
-open class EntityHookContext(
+abstract class BaseEntityExtensions(
```
czyzby, line 17: Consider not using plural form for class names. Describe the role of the provided extensions instead, e.g. `EntityHandler`, `ComponentContext`.

Quillraven:
@czyzby: I was able to use an abstract class for all the different areas - thanks again for the hint!

@jobe-m: I adjusted the `addOrUpdate` functionality as you suggested. Also, the `world.query` function and `EntityHookContext` class got removed because the world itself is now providing those extensions.

For that, a new `BaseEntityExtension` class got introduced to provide those basic extensions in different areas. Following classes inherit from it:
- Family
- IntervalSystem
- World
- EntityCreateContext

With this change, the hooks now take one less argument because the world is no longer passed. The hook itself runs within the world context which can be referenced via `this`.

Finally, I added a `notHas` infix function.

Now, instead of writing

```kotlin
world.query(entity) { it has Position }
```

you can write

```kotlin
with(world) { entity has Position }
```

src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
Quillraven: I already have an `EntityCreateContext` and `EntityUpdateContext`. What do you think about calling it `EntityBaseContext` or `EntityBaseComponentContext`?

Quillraven: I already have an `EntityCreateContext` and `EntityUpdateContext`. What do you think about calling it `EntityBaseContext` or `EntityBaseComponentContext`?

czyzby:
> `with(world) { entity notHas Position }`

It doesn't sound quite right, but `doesNotHave` is way too verbose for an infix such as this. Perhaps `hasNo` makes sense? `entity hasNo Position` sounds more or less like a sentence.

czyzby:
@Quillraven Just one more idea on how to approach `has`. You cannot add the `!` operator to the infix method itself, but there's nothing stopping you from overloading the operator for the companion object itself. Quick proof of concept:

```kotlin
typealias Entity = Int

abstract class ComponentType<T>(val componentClass: Class<T>) {
  // Add this operator:
  operator fun not() = Negated(this) // Perhaps cache this.
}

// Simple component type storage that marks it for negation:
data class Negated<T>(val componentType: ComponentType<T>)

// Example component:
class Position {
  companion object : ComponentType<Position>(Position::class.java)
}

// Relevant world methods:
class World {
  infix fun Entity.has(componentType: ComponentType<*>): Boolean = true // TODO
  // Add this:
  infix fun Entity.has(negated: Negated<*>): Boolean = !has(negated.componentType) 
}

// Usage example:
fun example() {
  with(World()) {
    val entity: Entity = 0
    entity has Position
    entity has !Position
  }
}
```

Whether `entity has !Position` is easier to understand than `entity hasNo Position`, I guess it's a personal preference thing.

Quillraven:
I will go for the `hasNo` version because I think it is easier to read from left to right. Your `Negated` version is a nice trick though! Did not know about that not operator :)

Quillraven:
Tested the new changes in Dinoleon and MysticWoods again, looks good! I like the new `with(world)` approach a lot more.

E.g. in the AIEntity class I mentioned above, I can now simply wrap the functions in the `with(world)` construct and can then intuitively work with the ECS:
image

Also the event handling example from above looks cleaner now imho:
image

Quillraven:
Added now also `Entity.configure` and `Entity.remove` to the `EntityComponentContext`. I guess this is everything that is needed. Here is an example from the test checking all the different functions:

image

jobe-m:
Awesome 👍 I will update my game tomorrow to the latest changes!

Quillraven:
Okay, wiki is updated. There are actually a few mistakes in the original documentation already (mainly code example typos) but I did not bother to fix them in the 1.x wiki as it will anyway die at some point.

In 2.0 everything should be correct :) Appreciate it, if anyone wants to read over it!

jobe-m:
Well done! 👍 💯 

I read through the wiki and also updated my game which still works flavlessly :)
One thing in the wiki I stumbed over. The following sentence on the "system_new" page is a bit hard to understand / follow:

```
The purpose of this function is performance reasons to trigger internal
calculations of Fleks only once instead of each time a component gets
added or removed.
```

With "this function" `entity.configure` is meant. But what does that mean? Currently it can be read also as "usage of the entity extention functions is also possible outside the configure block. Does this has performance penalties or does this work at all? (It shouldn't, I guess). Maybe you can rephrase it?

Quillraven:
Thanks for checking! I updated this section in the wiki and also updated the titles to remove the "_new". The updated section now is:

image

Quillraven:
I now removed the `kmp` branch. We have a history for 1.6-JVM and 1.6-KMP. Master is now up to date with Fleks2.

I will do some final checks tomorrow and then prepare the release.

Quillraven:
@czyzby / @jobe-m: What is a good name for our `addOrUpdate` function? Since we changed its implementation it does not make sense to me anymore. This function is now **always** calling update and only calling **add** if the component does not exist.
Therefore, `add OR update` is not correct imo. But also `updateAndAdd` is not correct? `updateAndAddIfNotExists` is way too long :D

Any good ideas?

```kotlin
/**
     * Updates a [component][Component] of the given [type] for the [entity][Entity] by calling [update].
     *
     * If the [entity][Entity] does not have such a [component][Component] yet then [add] is called to
     * assign the returned [component][Component] to the [entity][Entity] before calling [update].
     */
    inline fun <reified T : Component<T>> Entity.addOrUpdate(
        type: ComponentType<T>,
        add: () -> T,
        update: (T) -> Unit,
    ) 
```

mariorez:
maybe `refresh` ?

czyzby:
That's a tough one if you'd like to have this as a single method. I'd probably go with 2 separate methods along the lines of `addIfNull` and `update`, or simply a single `update` method with the `add` parameter being optional.

Quillraven:
Hmmmm 😅

Add is not really optional because it will be called if the component does not exist and also, the component instance will be created lazily via the lambda, because we don't want to create an instance that is not necessary.

Splitting it into two methods it will look like this:

```kotlin
entity.addIfNull{ Move() }.update{ cmp -> cmp.speed += 3f }


// or maybe if addIfNull simply returns either the new component or the existing one, we could do this
entity.addIfNull{ Move() }.speed += 3f

// for more complex components it can be combined with kotlin's 'also' or 'run' to do multiple things
```

I think that's maybe easier to just remove `addOrUpdate` completely and replace it with `addIfNull` ?

czyzby:
That's what I was proposing, yes. Maybe `getOrAdd` would be even better if you're going with a chainable method.

By the way, can you think of any way to define extension properties as a way to access and optionally create components? Something similar was recently added to `ktx-ashley`. It's tricky in your case, since entities are just type aliases, but I'm wondering if something along those lines would be cleaner.

```kotlin
import kotlin.reflect.KProperty

// Mock types:
typealias Entity = Int
interface Component

// Example component:
data class Position(var x: Float = 0f, var y: Float = 0f): Component

// Mock World implementation:
open class World {
  inline fun <reified T> Entity.has() = has(T::class.java)
  fun Entity.has(componentClass: Class<*>) = true
  fun <T> Entity.add(component: T): T = component
  fun <T> Entity.get(): T? = null
}

// Example implementation of property delegate:
class ComponentDelegate<T : Component>(
  private val world: World,
  private val componentType: Class<T>,
  private val provider: () -> T
) {
  operator fun getValue(entity: Entity, property: KProperty<*>): T = with(world) {
    if (entity.has(componentType)) entity.get<T>()!! else entity.add(provider())
  }

  operator fun setValue(entity: Entity, property: KProperty<*>, value: T) = with(world) {
    entity.add(value)
  }
}

// Factory method for property delegates:
inline fun <reified T: Component> World.propertyFor(noinline provider: () -> T) =
  ComponentDelegate(this, T::class.java, provider)

// Usage example #1 - global instance:
val world = World()
val Entity.position by world.propertyFor { Position() }

fun example() {
  val entity: Entity = 0
  entity.position.x = 10f

  // The problem is, it pollutes Int/original type API outside of world scope:
  val number: Int = 0
  number.position.x
}

// Usage example #2 - custom world extension:
class MyWorld : World() {
  val Entity.position by propertyFor { Position() }
}

fun usage() {
  val world = MyWorld()
  with(world) {
    // Only works on entities/ints within the world scope:
    val entity: Entity = 0
    entity.position.x = 10f
  }
}
```

Not sure if you can manage it without a global world instance or forcing the users to extend `World`, but this would probably be the most natural API you can get away with.

jobe-m:
Hmm, not sure if we should rewrite the whole API for fleks 2 again ;)  But it looks like an interesting approach to define components for an entity.

@Quillraven if you rename the `addOrUpdate` method I would keep the "update" in the name because actually I am using it mainly to update a component of an entity. If that entity does not have that component I would be ok if it is silently added.

But also the `getOrAdd` approach with the following `also` is for me ok and well readable. I even like `getOrAdd` more than the old `addOrUpdate` pattern.

Quillraven:
I actually checked out those commits a few weeks ago and also saw that `tagFor` functionality (I think it was called like that). I am not sure if we gain much with that given the new API because isn't the only difference than instead of writing `entity[Position]` you wirte `entity.position`? Or am I missing something?

I personally prefer the second option but I am not sure how easy it is to implement here. I could imagine your second example and combine it with the existing `world` DSL function and somehow use this given type of the user world instead of the fleks world internally to get access to the different properties. Something like:

```kotlin
// USE 'T' internally somehow instead of 'World'
fun <T : World> world(
    newWorld: T,
    // entityCapacity: Int = 512, <-- no longer necessary since the world is created outside
    cfg: WorldConfiguration.() -> Unit
): T {
    // !! NEW !! the line below becomes an argument to the function and we make World an open class
    // val newWorld = World(entityCapacity)
    CURRENT_WORLD = newWorld

    try {
        val worldCfg = WorldConfiguration(newWorld).apply(cfg)
        // assign world systems afterwards to resize the systems array only once to the correct size
        newWorld.systems = worldCfg.systems.toTypedArray()
    } finally {
        CURRENT_WORLD = null
    }

    return newWorld
}
```

Quillraven:
@jobe-m: I just updated the master with the new `getOrAdd` approach. I am currently migrating my three example games ;) Let me know, if this change is fine for you. After that I will release the version :)

I also don't see the property delegation approach for version 2 mainly because I don't have a lot of time anymore for the next couple of weeks/months and I want to get that update out with the version combining (JVM + KMP) and imo improvements over the old API.

Ideally, we can make the delegation approach, if feasible, as an update for 2.x and ideally it does not break anything meaning that users can stick to the current API but can also use the delegation properties if they prefer it.

czyzby:
> I actually checked out those commits a few weeks ago and also saw that `tagFor` functionality (I think it was called like that). I am not sure if we gain much with that given the new API because isn't the only difference instead of writing `entity[Position]` you wirte `entity.position`? Or am I missing something?

The extra utility is that you can define how the missing components are handled, and the user does not have to pass a lambda on every call. As for the performance - not sure how big of an overhead the properties are, but `getOrAdd` doesn't seem that lightweight itself with its lambda parameter. You'd have to compare the two.

Also, `tagFor` is for "boolean" components like `Visible` where the existence of the component itself is important info in terms of game logic. `propertyFor` works similarly to what I've posted above.

Anyway, this wouldn't require a rewrite of the API, since the properties could be built upon the existing one. It could be an optional, complementary syntax to the existing `entity[Position]`. The only required change  make it work is that you'd probably have to make the `World` extensible so that the users can define their own `Entity` extensions within.

Issue #71, opened 2022-09-29
Unexpected(?) behavior with nested `entity {}` calls


rubybrowncoat:
This is an issue I also encountered with 1.6 but it was just before starting the update to 2.0 and I was hoping it wouldn't happen there.

```kotllin
var playerEntity: Entity?
var subEntity: Entity?

val entityWorld = world(entityCapacity = 1000) {
    // World things
}

with(entityWorld) {
    playerEntity = entity {
        it += Player()
        it += Storage().apply {
            subEntity = entity {
                it += Identifier("SUPPLIES")
                it += Quantity(210)
            }

            contents += subEntity!!
        }
        it += StageActor()
        it += TextureAnimation()
        it += PhysicsBody()
    }
}

entityWorld.snapshotOf(playerEntity!!)
entityWorld.snapshotOf(subEntity!!)
```

The internal details of PhysicsBody, TextureAnimation, StageActor and Player shouldn't be relevant. Storage has one property `contents` (a MutableList of Entity). 

The issue is that `snapshotOf(playerEntity!!)` shows it having only the Player() component, but it should have Storage, StageActor, TextureAnimation and PhysicsBody as well. While `snapshotOf(subEntity!!)` crashes with this error: 

```
Exception in thread "main" com.github.quillraven.fleks.FleksNoSuchEntityComponentException: Entity 'Entity(id=1)' has no component of type 'TextureAnimation'.
	at com.github.quillraven.fleks.ComponentsHolder.get(component.kt:131)
	at com.github.quillraven.fleks.World.snapshotOf(world.kt:448)
```

Quillraven:
Nesting is not allowed and is at the moment intentional. Why can you not create your subentity before?

You will get a similar problem if you nest e.g. entity.configure calls with different entities.

The reason is that the EntityCreateContext gets the component mask of the newly created entity set before running your entity block code that adds the components. With your nesting you then override that with a the mask of the second entity and your remaining calls for the first entity get applied to the mask of the second entity.

We can think about changing that but I never saw someone nesting entities before. What is your reason for doing that?

rubybrowncoat:
I'm not sure there's an intentional reason, it just felt natural to create the entities that go into a new Storage component of a new entity while creating the entity and the component themselves, since I don't need other references to them outside of that storage component's contents property (the subEntity variable above is just for example's sake). Storing them in variables and then adding them to the component just seems like an extra step, but shouldn't be an issue.

[...]

PR #72, opened 2022-09-30
Add support for nested create/configure calls

Quillraven:
As mentioned in #71, it would be nice to allow nested entity creation. To do that we need to slightly adjust the `entity.kt` file to always directly access the component mask of the entity in the create/update context instead of setting it beforehand.

I made those adjustments and checked if this has any impact on the performance and it doesn't seem like it. So the change seems to be fine and avoids the issue with nested calls for both `create` and `configure`.

Test cases are of course added. They failed before but now they are successful!

----

@rubybrowncoat: Can you please try those changes locally in your game? To do that, just check out this branch locally and execute the `publishToMavenLocal` gradle task **after** changing the Fleks version to e.g. 2.1-SNAPSHOT. The important part is the `-SNAPSHOT` so that you don't need to sign the created files.

image

----

I won't release a 2.1 version so fast because there might be other things popping up in the next 2 or 3 weeks but you can work with this updated version already locally until then.

### DIFF
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
@@ -1,9 +1,6 @@
 package com.github.quillraven.fleks
 
-import com.github.quillraven.fleks.collection.BitArray
-import com.github.quillraven.fleks.collection.IntBag
-import com.github.quillraven.fleks.collection.bag
-import com.github.quillraven.fleks.collection.compareEntity
+import com.github.quillraven.fleks.collection.*
 
 /**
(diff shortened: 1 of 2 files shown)


rubybrowncoat:
I have tried the branch and can confirm that the nesting `entity {}` calls is working correctly. I'll report back if I notice anything unexpected happening related to this enhancement.

Thanks!

Quillraven:
@rubybrowncoat: I merged this PR and will close the issue. I will prepare a new version 2.1 in two or three weeks including also an upgrade to Kotlin 1.7.20 and maybe other things.

If you face other issues/questions, please feel free to raise them :)

Thank you for your feedback!

PR #74, opened 2022-10-12
Collection functions for entities

Quillraven:
As per the #66 discussion, this PR is updating the existing `IntBag` functionality with certain Kotlin collection stdlib functions.

Following things are adjusted:
- rename `IntBag` to `EntityBag`: this change also includes converting the underlying `IntArray` to `Array<Entity>` **and** changing `Entity` back to a value class. For some reason it is a data class (I guess this is some leftover of the previous KMP branch changes)
- rename `entityBag` inside family to `entities` and expose it (internal -> public)
- add a new `asEntityBag` function to the world
- add collection stdlib functions to `EntityBag`

Additionally, this PR updates Kotlin and Dokka from 1.7.10 to 1.7.20.

Code adjustments are done but I am missing the tests currently and the documentation of the new functions. Following functions are now available for `EntityBag`:
- `size`: this is actually a property
- `contains`
- `containsAll`
- `isEmpty`, `isNotEmpty`
- `all`, `any`, `none`
- `associate`, `associateBy`, `associateTo`, `associateByTo`
- `count`
- `filter`, `filterNot`, `filterIndexed`, `filterTo`, `filterNotTo`, `filterIndexedTo`
- `find`
- `first`, `firstOrNull`
- `fold`, `foldIndexed`
- `forEach`, `forEachIndexed`
- `map`, `mapIndexed`, `mapTo`, `mapIndexedTo`
- `random`, `randomOrNull`
- `take`


rubybrowncoat:
Ran a few tests with my local version, filtered some families and ran a couple of folds and maps to see if things were working. It looks ok.

I also tried accessing a component of an entity inside the component of another entity and that seemed to work as well, something like this: 

```kotlin
val totalTimberInAllMarkets = family.entities.fold(0) { totalAccumulator, familyEntity -> 
    totalAccumulator + familyEntity[Storage].contents.filter { filterEntity -> 
        filterEntity.getOrNull(Identifier)?.identifier == "TIMBER" 
    }.fold(0) { storageAccumulator, storageEntity -> 
        storageAccumulator + (storageEntity.getOrNull(Quantity)?.quantity ?: 0) 
    } 
}
```

While trying this a question came to me: do you see us using `EntityBag` for entity lists or you think it's a stricly fleks-internal structure? I ask this because my `Storage` component has a contents property that is just a mutable list of `Entity`. Should I be using EntityBag instead?

Quillraven:
I thought about the `EntityBag` question as well but I see one problem. The modification methods like add/remove are internal because otherwise users couuld potentially add/remove entities to families which is forbidden because it bypasses any Fleks internal logic.

I then thought of having two implementations: `ImmutableEntityBag` and `MutableEntityBag`. The mutable version could be exposed by the new methods instead of list. However, I am not sure if there is a big advantage over it because the bag internally uses an array which needs to be resized to fit in all necessary entities. Especially with a filter functionality we don't know in advance how many entities need to be added and therefore a few resizes could happen which is not ideal.

Not sure how Kotlin's list is implemented to be honest. Maybe it has the same problem?

Quillraven:
I added now the tests and actually had to adjust some of the methods because they were wrong ;) But not the ones you were using.

One more question regarding your storage. Do you often iterate over those entities? And do you know the average maximum of entities that is stored in the Storage in advance? If both questions can be answered with yes, then EntityBag should be better.

I will change the internal implementation a little bit to follow the ImmutableEntityBag and MutableEntityBag approach. Then people can use the Mutable version if they want.

Quillraven:
@rubybrowncoat Can you try it once more please? You can now use `MutableEntityBag` in your code if you want. It provides add/remove functionality plus all the functions mentioned above.

For the filter functions I now assume that ~25% of the original bag size will be needed. I am not sure which value should be used here ideally. If more entities are part of the result then some array resizing is happening. If less entities are needed then a little bit of memory space is wasted. However, iteration and other followup chaining calls should be better.

rubybrowncoat:
I've given it a test, the operations I was running before are still working correctly. I have seen no changes in results with the new version, which is what I expected.

While trying a couple of other things I found myself looking for a method that wasn't implemented (`mapNotNull`) while doing `family.entities.map { it.getOrNull(Market)?.parentMarket }`. Market is a component that might not be on the entity and parentMarket is a property that can optionally contain a Market component.
I could easily get to the desired end result by concatenating `.filterNotNull()` after the map. I don't know what the internals of `mapNotNull` usually are, if it's just an alias for `.map.filterNotNull` or if it is slightly more efficient at it.

In the same vein, while trying things, I found myself wanting to do a `.flatMap` (would have liked to do a flatMapNotNull, to be precise, even if that doesn't usually exist on standard collections). I would have done something like this `family.entities.flatMap { it.getOrNulll(Storage)?.contents ?: emptyEntityBag() }`, or `family.entities.flatMapNotNull { it.getOrNulll(Storage)?.contents }`, to get all the entities in all the storages.

I've also thought about doing a `.groupBy`, which would ideally make a Map of EntityBags.

---
Unrelated to the collection functions, a floating question: can Components inject from the world the way Systems can? If they can't, would it make sense if they did?

```kotlin
class CombatPhysicsBodySystem(
    combatManager: CombatManager = World.inject(),
) : IteratingSystem
```

Quillraven:
`mapNotNull` and `mapNotNullTo` make sense. My initial thought was that an `EntityBag` never contains null values and therefore the `...NotNull` methods don't make sense. But actually in case of map it does make sense :) I will add them.

Same for `flatMap` and `groupBy`. I can add them as well of course.

----

For your injection question for components: no, that doesn't work and imo should also not work because components are just a "stupid bag of data". They don't know about anything. They just hold some information and should not care about any ECS specifics like a world.

However, normally when you add components to an entity then you also have your world instance available. In this case you can of course pass your injectable to the component constructor. Something like:

```kotlin
class MyComponent(val someInjectable: InjectableType) : Component {
    // ...
}

// somewhere in your ECS code
world.entity {
    it += MyComponent( world.inject() )
}
```

Quillraven:
I have added now following functions in addition:
- `flatMap`, `flatMapSequence`, `flatMapBag`, `flatMapNotNull`, `flatMapSequenceNotNull`, `flatMapBagNotNull`: for some reason I had to give them unique names because otherwise I got weird compiler errors. Iterable/Sequence have the same JVM signature but adding @JvmName to one of the methods and giving it a unique name didn't help. And the other problem was that the compiler seemed to be confused if it should use the iterable version or EntityBag version of the method. I don't know how to fix that, that's why I gave them unique names for now. If anyone knows how to solve that then please let me know!
- `groupBy`, `groupByTo`:  I saw there is also a `groupingBy` but I don't know how to create a grouping instance and if we need it.
- `mapNotNull`, `mapNotNullTo`

Issue #76, opened 2022-11-22
2.1 family doesn't get updated


ernespn:
Hi Simon,
I think I found a bug in the 2.1 version.
So I have a extension that check if a certain family is in a position:
``` kotlin
fun Position.findEntitiesInPosition(family: Family): List<Entity> {
    val entities = mutableListOf<Entity>()
    family.forEach {
        if (it.has(Position)
            && it[Position].x == x
            && it[Position].y == y
        ) {
            entities.add(it)
        }
    }
    return entities
}
```
I use this to see if I can move to a target position
``` kotlin
class MoveSystem(
    private val maxWidth:Int = inject("maxWidth"),
    private val maxHeight:Int = inject("maxHeight")
) : IteratingSystem(family { all(Position, Moving).none(Dead)}) {
    override fun onTickEntity(entity: Entity) {
        val originalPosition = entity[Position]
        val targetPosition = entity[Moving].target
        // position is within the limit and position is not occupied
        if (targetPosition.positionWithin(maxWidth, maxHeight)
            && targetPosition.findEntitiesInPosition(world.family { all(Position, Solid) }).isEmpty()){
            originalPosition.x = targetPosition.x
            originalPosition.y = targetPosition.y
        }

        entity.configure { it -= Moving }
    }

```
So for the version 2.0 this was working fine but I recently upgraded to 2.1 and now it is failing the test below:

``` kotlin
    @Test
    fun shouldMoveIntoADeadObject(){
        val x = 5
        val y = 5
        val original = Position(x, y)
        val target = Position(8,8)
        val entity = world.entity {
            it += original
            it += Moving(target)
        }

        val toDie = world.entity{
            it += target
            it += Solid()
        }

        world.update(1f)
        // Check I cannot move
        Assertions.assertEquals(original, Position(x,y))
        Assertions.assertNotEquals(original, target)

        with(world){
            toDie.configure {
                it -= Solid
                it += Dead()
            }
            entity.configure { it += Moving(target) }
        }

        world.update(1f)
        // Check it should have moved
        Assertions.assertEquals(original, target)
    }

    
```

Quillraven:
Hm I wouldn't know which change was introduced in 2.1 that could cause that 😅 weird.

Anyway, can you let me know please which assertion exactly is failing? Also, which family doesn't get updated? The one from the system or the one that you create inside the onTickEntity method?

[...]

PR #77, opened 2022-11-23
#76 family not updated

Quillraven:
Fixes a bug that families did not get updated properly when used in a nested way (#76).
I still need to do the final tests on my PC with my projects including a performance run. I hope adding an additional 'isIterating' field isn't that critical ;)

Also, updated Kotlin from 1.7.20 to 1.7.21 and kotlinx.benchmark to 0.4.5.

@ernespn: can you try this version in your project as well please? Just check out this branch, add "-SNAPSHOT" to the version in the gradle file and execute the publishToMavenLocal task to create version 2.1a-SNAPSHOT on your computer.

If you need further  help setting this up please let me know.

And thank you for reporting this issue!

### DIFF
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -1,10 +1,10 @@
 # Fleks
 
 [![MIT license](https://img.shields.io/badge/License-MIT-blue.svg)](https://github.com/Quillraven/Fleks/blob/master/LICENSE)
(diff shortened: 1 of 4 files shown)


ernespn:
Yes, the fix works!
Thanks for fixing it so quickly :)

Quillraven:
I will release an official version on the weekend if nothing goes wrong. Again, thanks for bringing this issue up and for testing the fix in your project! 

Let me know if it is ready in some state then I can add it to the example projects in the wiki :)

Issue #78, opened 2022-11-24
`EntityBag` interface does not have a `get` operator function


ScottHamper:
Hey Quillraven,

While prototyping a naive collision system, I've discovered that the `get` operator function on `MutableEntityBag` is not part of the `EntityBag` interface. Is this intentional?

My current approach is to create an `IntervalSystem` that manually iterates over a family, in order to avoid duplicating collision checks between entities. Something like the following:

```kotlin
class CollisionSystem(interval: Interval) : IntervalSystem(interval) {
    private val family = world.family { all(PositionComponent, HitcircleComponent) }

    override fun onTick() {
        for (i in 0 until family.numEntities) {
            for (j in i + 1 until family.numEntities) {
                val first = family.entities[i] // Error; `EntityBag` interface does not provide a direct way to retrieve an item
                val second = family.entities[j] // Error

                // Check for a collision between `first` and `second`
            }
        }
    }
}
```

Since the `get` operator does not mutate the `EntityBag`, should it be added to the `EntityBag` interface rather than be an addition specific to `MutableEntityBag`? If there's a reason to leave it out of `EntityBag`, do you have any suggestions for how I might prevent a collision system from checking the same pair of entities more than once? This is my first foray into using an ECS, so perhaps I am approaching the situation in a less than ideal way.

Quillraven:
I think there is no harm in adding it. As you mentioned it does not mutate anything so it is fine.

The reason it is missing that so far it was nit needed. In most cases you iterate with forEach and accessing a specific index in the bag was not necessary for users imo.

However in your case it makes sense. For a quick and naive system I think this is fine but your approach is of course not optimal because you need to iterate over the entities twice and the bigger the bag the longer your iteration takes.

When it comes to collision it is also normally not necessary to check everything with everything because a lot of entities will not even be in range if each other and the check is therefore wasted.

In a 2D environment this is sometimes solved via a so called Quadtree structure to optimize it.

Said that, if you don't have thousands of entities I think your approach is fine for now.

I haven't done collision handling myself for a long time because I always use Box2D lately in my projects and that already does the collision handling for me 😉

ScottHamper:
> When it comes to collision it is also normally not necessary to check everything with everything because a lot of entities will not even be in range of each other and the check is therefore wasted.
>
> In a 2D environment this is sometimes solved via a so called Quadtree structure to optimize it.

Yea, for sure! Quadtrees are great.

[...]

PR #79, opened 2022-11-24
Add `get` operator function to `EntityBag`

ScottHamper:
Also add unit test to assert that `get` throws an `IndexOutOfBoundsException` when its `index` parameter is negative.

Fixes issue #78.

No worries if this change is undesired! Putting the pull request together only took a few minutes - just hoping to save you some time if it's a change you do want.

I moved the `get` function further down in `MutableEntityBag`. It seemed like the class layout included functions specific to `MutableEntityBag` first, followed by the `EntityBag` interface functions in alphabetical order.


Quillraven:
Wow cool, thank you for the contribution! Looks good to me 👍

I try to release a new version over the weekend which contains a Bugfix as mentioned in #77 

I will also include your change there.

ScottHamper:
Awesome! Thanks for such a timely response.

Issue #80, opened 2023-01-21
Partial world update by network


Gidroshvandel:
Hello, I am trying to make a network game using fleks.

I only found the snapshot api for mass updating the world. But I don't want to transmit the whole world over the network. It would be convenient to be able to add entities received over the network with their components.

Could you tell me the best way to implement this by using fleks?

Quillraven:
Hi,

Could you give me more details what you are trying to achieve? Maybe also a code example how you want your code to look like?

Because I am not 100% sure what your problem is. 
The snapshot functionality is meant for saving/loading an entire world. Typical use-case is a save and load system in a game.

Other entity creation while the game is running should be done with world.entity.

I normally have helper functions for different entity types. E.g. one function to create a NPC with it's necessary components. One function for a special effect entity, etc.. Those functions then just take a minimum amount of arguments like e.g. NPC would be XY coordinate and NPC type. The function then creates an entity with a position, sprite, animation, AI, ... component.


Maybe this helps you already? Otherwise, please provide more details about your issue and how you'd like to solve it.

If something is missing I am gladly adding it to the library 🙂

Gidroshvandel:
For example i have class like this to track changes in the world
```
sealed class UpdateData() {
    class Changed(val entityListMutableMap: Map<Entity, List<Component<*>>>) : UpdateData()

    class Added(val entity: Entity, val list: List<Component<*>>) : UpdateData()

    class Removed(val entity: Entity) : UpdateData()
}
```

And i want to apply this changes

```
fun applyUpdateData(list: List<UpdateData>) {
        list.forEach { updateData ->
            when (updateData) {
                is UpdateData.Added -> world.entity {
                    if (it.id != updateData.entity.id) { // I'm not sure that the entity will be created with the correct id
                        throw IllegalStateException()
                    } else {
                        updateData.list.forEach { component ->
                            it += component // Entity.plusAssign required reified T : Component<T> but my component is Component<*>
                        }
                    }
                }
                is UpdateData.Changed -> {
                    updateData.entityListMutableMap.forEach { entry ->
                        entry.key.configure {
                            entry.value.forEach { component ->
                                it += component // Entity.plusAssign required reified T : Component<T> but my component is Component<*>
                            }
                        }
                    }
                }
                is UpdateData.Removed -> updateData.entity.configure {
                    it.remove()
                }
            }
        }
    }
```

Perhaps I chose the wrong way to solve the problem, can you tell me the best way to do this?

[...]

PR #81, opened 2023-01-22
Networking functionality ideas

Quillraven:
PR for #80 

I added a new `loadSnapshotOf` function to the world which allows to set components for a specific entity. If the entity already exists, then its components get replaced with the new ones. If the entity does not exist, it will be created before setting its components.

@Gidroshvandel, @jobe-m : I'd like to continue the discussion in this PR please. This is not final yet. I suggest that we wait for the feedback of @Gidroshvandel and his network game before merging anything of this branch.

Imo this function could still be useful for Fleks in general but let's see what kind of other things we might add for better networking support ;)

### DIFF
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
@@ -248,16 +248,30 @@ class EntityService(
     /**
      * Updates an [entity] with the given [components].
      * Notifies all [families][World.allFamilies].
-     * This function is only used by [World.loadSnapshot] and is therefore working
-     * with unsafe wildcards ('*').
+     * This function is only used by [World.loadSnapshot] and [World.loadSnapshotOf],
+     * and is therefore working with unsafe wildcards ('*').
      */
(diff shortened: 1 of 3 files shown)


Quillraven:
@Gidroshvandel: do you know how to checkout this branch and build it for your mavenLocal repository so that you can use it in your game?

You need to adjust the verrsion in the `build.gradle.kts` file and make sure that it ends with "-SNAPSHOT". E.g. "2.3-SNAPSHOT". The "-SNAPSHOT" makes sure that signing is not happening and then you can run the `publishToMavenLocal` task.

If you need more help, just let me know!

Gidroshvandel:
Thank you! I'll check it out soon and post the results.

jobe-m:
I agree that "reserving" IDs for a world is not pratical. I like the idea of the NetworkComponent holding a mapping of local and remote IDs.

Quillraven:
> Thank you! I'll check it out soon and post the results.

@Gidroshvandel any feedback so far? Did you have some time to try it out? :)

Quillraven:
@Gidroshvandel I will merge this PR next weekend with some other minor things and release a new version.

Would be still great to get your feedback!

Issue #83, opened 2023-02-16
More dynamic world and entity building


chrispytoes:
My game is heavily based on large amounts of content, and as such, I'm using a combination of config files and a SQLite database to manage it all. On top of that, I also want my game to have good mod support, so I really need a way to do things dynamically like adding component `onAdd` and `onRemove` listeners, or adding components to entities without access to their exact types.

For instance, this doesn't work, as it complains that not enough type information is given:
```
abstract class GameComponent<T>: Component<T> {
    abstract fun onComponentAdded(entity: Entity)
    abstract fun onComponentRemoved()
}

private val gameComponents: MutableList<GameComponent> = mutableListOf()

world {
  components {
    gameComponents.forEach {
      onAdd(it.type()) { entity, component ->
        it.onComponentAdded(entity)
      }
      onRemove(it.type()) { entity, component ->
        it.onComponentRemoved()
      }
    }
  }
}
```

Or this:
```
val someComponents: MutableList<Component<*>> = mutableListOf()
        
world.entity { entity ->
    someComponents.forEach {
        entity += it
    }
}
```

How big of a breaking change would that be to make things like this possible? I basically want to build my own "world builder" object that I can pass around to various places that add new features to the world, and then build it all at once and start it. The entity building problem I can solve by adding an abstract method to `GameComponent` to delegate adding the component, but I'm stuck on adding component listeners dynamically.

Quillraven:
Hi,

can you have a look at the currently open PR and the new `loadSnapshotOf` function? That might solve yourr problem with adding components dynamically without access to their exact types.

About your componen onAdd/onRemove question:

I am not sure if I like to add that directly to our current world  builder because in my opinion you lose the type-safety because you cannot really use reified types in your approach, right? 
I also have a question for you because in 2.x I removed the listener approached and changed it to "hooks" and with that change I also only allowed one hook per component type because it made sense for the - imo - typical use-case for special init/cleanup logic for a component (like Box2D). But if I understand you correctly, you need more than one hook per component?

I have a different proposal, please let me know about your feedback. In my games I usually have special functions to spawn entities and since Fleks 2.x you really add component instances to your entity instead of types that are then created via reflection. That way it is up to the user how he wants to create those instances and e.g. make pooling for some components while doing the normal creation approach for other components.

In your case I could imagine that you create your own `ComponentFactory` or whatever you want to call it, add it as a dependency to your world and use it for component creation/removal.

[...]

PR #84, opened 2023-02-22
Small additions for 2.3

Quillraven:
- update gradle to 7.6
- minor grammar fixes
- new `componentTypeOf` function
- new `plusAssign` with `List<Component<*>>` argument. Might also be useful for exceptional cases like #83 and #80 

@jobe-m : I know it is nothing huge but would still appreciate your feedback for the new plusAssign. Could it be useful? Or shall we remove it?

Also, I am thinking of introducing an `EntityHook` similar to the `ComponentHook` that we have. It will trigger `onAdd` after the components are added to a newly created entity. It will trigger `onRemove` before components are removed for a removed entity.

I recently had one or two cases where my first guess was to use an `onRemove` event for entities and saw that currently we don`t have that anymore. I was able to solve my cases differently and imo better but I could imagine that it could still be a useful addition to the library. What do you think?

----

edit: by the way the `componentTypeOf` function can be more convenient for components with more than one type. I had a real example now in one of my games where I have a `Light` component for box2d-lights and an entity can have multiple lights but only one of each type like point or cone.

Of course you can solve that by storing a list of lights in the component but in that specific case I preferred to have different component types. With this addition I can avoid the ugly `object : ComponentType<Light>() {}` syntax and change it to `componentTypeOf<Light>()`.

Here is the component I am talking about:
```kotlin
data class Light(
    val light: B2DLight,
    val distance: ClosedFloatingPointRange<Float>,
    val angle: ClosedFloatingPointRange<Float> = 0f..0f,
    var distanceTime: Float = 0f,
    var distanceDirection: Int = -1,
) : Component<Light> {

    private val cmpType = if (light is PointLight) LightPoint else LightCone

    override fun type(): ComponentType<Light> = cmpType

    companion object {
        val LightPoint = object : ComponentType<Light>() {}
        val LightCone = object : ComponentType<Light>() {}

        private const val numRays = 64
        val ambientColor = Color(0.05f, 0.05f, 0.05f, 1f)
        const val ambientOrbGain = 0.03f
        val distanceInterpolation: Interpolation = Interpolation.smoother
        val angleInterpolation: Interpolation = Interpolation.swing

        val onRemove: ComponentHook<Light> = { _, component: Light ->
            component.light.remove()
        }

        fun pointLightOf(
            rayHandler: RayHandler,
            color: Color,
            distance: ClosedFloatingPointRange<Float>,
            boundary: Boundary,
            body: Body
        ): Light {
            val (x, y) = boundary

            return Light(
                PointLight(rayHandler, numRays, color, distance.endInclusive, x, y).apply {
                    attachToBody(body)
                    // softness length allows the light to go through objects.
                    // E.g. this makes the player light illuminate trees when he is close to them.
                    setSoftnessLength(3.5f)
                },
                distance
            )
        }

        fun coneLightOf(
            rayHandler: RayHandler,
            color: Color,
            distance: ClosedFloatingPointRange<Float>,
            angle: ClosedFloatingPointRange<Float>,
            direction: Float,
            boundary: Boundary,
            body: Body
        ): Light {
            val (x, y, w, h) = boundary

            return Light(
                ConeLight(rayHandler, numRays, color, distance.endInclusive, x, y, 0f, angle.endInclusive).apply {
                    attachToBody(body, w * 0.5f, h, direction)
                    setSoftnessLength(3.5f)
                },
                distance,
                angle
            )
        }
    }
}
```


src/commonMain/kotlin/com/github/quillraven/fleks/component.kt
```diff
@@ -26,6 +26,12 @@ abstract class ComponentType<T> {
     }
 }
 
+/**
+ * Function to create an object for a [ComponentType] of type T.
+ * This is a convenience function for [components][Component] that have more than one [ComponentType].
+ */
+inline fun <reified T> componentTypeOf(): ComponentType<T> = object : ComponentType<T>() {}
```
jobe-m, line 33: Nice! Will come in handy also here ;)

image

src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
```diff
@@ -97,6 +97,27 @@ open class EntityCreateContext(
...
+     * It is preferred to use the [plusAssign] function whenever possible to have type-safety.
+     *
+     * If a component [addHook][ComponentsHolder.addHook] is defined then it
+     * gets called after the component is assigned to the [entity][Entity].
+     *
+     * If a component [removeHook][ComponentsHolder.removeHook] is defined and the [entity][Entity]
+     * already had such a component then it gets called with the previous assigned component before
+     * the [addHook][ComponentsHolder.addHook] is called.
+     */
+    inline operator fun Entity.plusAssign(components: List<Component<*>>) {
```
jobe-m, line 113: I think this will be useful. Since I am currently playing with serialization of entity and components. It will be very useful if one want to update "unspecific" amount of components of an entity. This will just be easier than if in that case you need to check for each component type by yourself.

/offtopic
BTW I tried again Kotlinx.serialization with a world snapshot and I got a bit further. I need to "re-map" the map from the snapshot, but I got around the "Any" and the Star projection problem :)

```json
{
    "entity": "42",
    "components": [
        {
            "type": "TestComponent",
            "payload": {
                "x": 42.1,
                "y": 17.3
            }
        }
    ]
}
```

Quillraven: Anything we can do to improve that to maybe avoid the re-mapping?

jobe-m: It is OK for now and I won't change the map of snapshot().

I don't know if I will change the structure again when I find an easier setup which can be serialized. The problem is not the entity as key, but the list of `Component<*>`. First I need to cast it to `Component<Any>`. For `Component<Any>` I have written a KSerializer strategy. "payload" is of type `Componenty<Any>` but it lost its underlying type. So I need that from "type" as string. In the `KSerializer<Component<Any>>` the string is than mapped to the actual (default) serializer of the final component (here: `TestComponent(...) : Component<Any> {...}`. This is totally weird but it works.
 
So -after thinking about it- we could change `snapshot(): Map<Entity, List<Component<*>>>` to `snapshot(): Map<Entity, List<Component<Any>>>` -> `*` to `Any` :)
I will continue to look into my code and maybe I find a way to directly serialize the returned map.

Quillraven: I just read your edit now AFTER I adapted the functions haha.

I had to make it Component<out Any> instead of Component<Any> because otherwise some type interference did not work properly.

But the change is no longer needed if I understand you correctly? Then I'd prefer to revert it.

jobe-m: Yes, you can revert it. I found a serializer solution which is in principle totally independent of Component<*> type. I will update my korge-fleks repo the next days for anyone who is interested:)

PR #86, opened 2023-03-05
Add kotlinx serialization support to fleks

jobe-m:
To be able to serialize world snapshot of fleks the kotlinx serializer
needs to know "how" to serialize Entity value class. Thus, it would
be nice to add @Serializable annotation from kotlinx.serialization
directly to the Entity class declaration. With that it is possible to serialize
the world snapshot (or snapshotOf) in a kotlin multiplatform game that
uses kotlinx.serialization.

In korge-fleks there are now unit tests which check serialization of
components. :-)
This check currently fails to to the missing addition in fleks: 
https://github.com/korlibs/korge-fleks/pull/4

No need to create directly new release in Maven. I can use the commit
hash or a local pre-release tag with KorGE kproject
<https://github.com/korlibs/korge-fleks/blob/main/example/libs/fleks.kproject.yml>.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,6 +4,7 @@ plugins {
     kotlin("multiplatform") version "1.7.21"
     id("org.jetbrains.kotlinx.benchmark") version "0.4.5"
     id("org.jetbrains.dokka") version "1.7.20"
+    id("org.jetbrains.kotlin.plugin.serialization") version "1.4.0"
     `maven-publish`
     signing
 }
@@ -55,7 +56,11 @@ kotlin {
     }
 
     sourceSets {
-        val commonMain by getting { }
+        val commonMain by getting {
+            dependencies {
+                implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:1.4.0")
+            }
+        }
         val commonTest by getting {
             dependencies {
                 implementation(kotlin("test"))
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
@@ -1,12 +1,14 @@
 package com.github.quillraven.fleks
 
 import com.github.quillraven.fleks.collection.*
+import kotlinx.serialization.Serializable
 import kotlin.jvm.JvmInline
 
 /**
  * An entity of a [world][World]. It represents a unique id.
  */
 @JvmInline
+@Serializable
 value class Entity(val id: Int)
 
 /**


Quillraven:
Hm but this adds kotlinx serialization as a dependency if I see that correctly? Meaning the library size gets bigger? By how much?

Also, it will add a strong coupling between Fleks and kotlinx serialization. If people want to serialize differently they will still get kotlinx in their project with this change?

I am not a serialization expert but those are the things that come to my mind. Is kotlinx serialization the standard? Then it is not that big of a deal.

Also, is it possible to maybe make the dependency optional?

Maybe we should allow people to specify the type of entity they want to use in a world? As default it will be the current Fleks entity and EntityService but it can be overruled by users.

Since you don't need a maven release right now, should we just create a separate serialization branch with this commit? Would that help you?

jobe-m:
You are right. Currently I don't know the full impact this has on fleks. Probably it's not much since the serialization plug-in adds some code to return an Int when serializing Entity.id. But I don't know if there is more added. In bad case the project which uses fleks world include the kotlinx.serialization lib. Even when it does not use serialization. At least kotlinx.serialization looks to be the standard serialization in Kotlin.

Would it be possible to change value class Entity to interface Entity? Then we do not need to annotate Entity with @serializable since interfaces are serializable by default. And fleks wouldn't depend on kotlinx.serialization. The project would then need to implement that interface e.g. as value class Entity. But don't know if that would work internally in fleks...

Quillraven:
I mean for now I don't have a problem with creating a separate "serialization" branch with your commit so you can use it, if that helps?

I will have a deeper look into it in April, where I should have more time again :)

My first idea was to check if a dependency is part of the classpath then use implementation X, otherwise implementation Y. That way I wanted to distinguish between the serializable entity and the normal one. But from a quick research this seems to require reflection which is not ideal and not sure if it even works in KMP.

Your interface approach might work but not sure if it has big downsides to performance. Would need to check that. The value class, from what I understood, compiles to a normal int. Not sure what the compiler does when it is an interface. Will also check that when I have time.

Anyway, for now, is a separate branch okay for you?

jobe-m:
Yes, of course a separate branch is OK for now.
I also thought a bit about Entity as an interface and I think it won't be feasible in a way which does not decrease performance. Maybe it makes most sense to have two implementations which could be selected somehow when needed. But we should check first what serialization is really adding. Maybe it is not worth the effort when serialization is adding just some Bytes/KBytes...

Would be nice to have a tag on that branch showing the related base release. E.g. 2.3.1 or 2.3-korgefleks1 or similar. No maven release needed :)

jobe-m:
What just came into my mind: We could change `value class Entity(id: Int)` to `typealias Entity = Int`? The breaking change would be that outside of Fleks an object like `entity` would be an Int. So all code would need to change from `entity.id` to just `entitiy`. The good thing is that the namespace `Entity` still exists with its nice extention functions. 

Since Int is a primitive type we would not need to annotate Fleks with kotlinx.serialization. I will try this out the next days and let you know. We can keep the PR open for now. I am in no rush :)

Quillraven:
I think this also has the downside that normal Int also get the extensions, but not 100% sure. But definitely a good idea and worth trying 😉

Btw, I will create the new branch in the evening and also create a new issue to write down the ideas about serialization and to not forget about this topic for a future release!

jobe-m:
Yes, you are right again :) Now I remember the discussion about polluting the Int namespace with extentions. Actually it would be just too elegant to have Entity as typealias ;)

PR #89, opened 2023-05-17
Update versions and Java compatibility from 8 to 11

Quillraven:
update versions:
* Kotlin 1.8.21
* dokka 1.8.10
* benchmark 0.4.8
* Gradle 7.6.1

Also, change the Java compatibility from 8 to 11 because I think it can give some benefits and also the reason I set to 8 initially was for Android. I think almost all relevant Android devices nowadays can handle Java 11 so it should not be an issue.

Any concerns regarding these changes? ;)

For the benchmark stuff I had to update the build file a little bit because otherwise it wouldn't work with Kotlin 1.8. That was also the reason why I did not update to 1.8 in version 2.3 because I didn't know how to fix it.

But I found a workaround for now which seems to work!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,12 +22,13 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
-      - name: Setup JDK 8
-        uses: actions/setup-java@v1
+      - name: Setup JDK 11
+        uses: actions/setup-java@v3
         with:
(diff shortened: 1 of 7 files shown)


jobe-m:
Looks good to me.
Regarding the version updates: Using Fleks in KorGE is not affected by it. Korge-fleks uses kproject to include Fleks as sources. Thus the used versions of Kotlin, Java, Gradle plugins, etc. are defined anyway by KorGE.
BTW since KorGE 4.0 we use already Java 11 which works fine already with Fleks.

Issue #87, opened 2023-03-06
Better serialization support (KorGE)


Quillraven:
As discussed in #86 it seems to be problematic that the Fleks entity does not have the `@Serialization` annotation which makes it more complicated on user side to serialize/deserialize the snapshots of a world.

The commit of #86 is part of a new branch: `2.3-korge-serialization`

----

Things that we need to consider:
- is kotlinx-serialization the standard to serialize/deserialize stuff? Or are there other libraries out there that people might use.
  - What is LibGDX using? I know that I was able to simply convert the snapshot to a  JSON string in LibGDX and store that in a preference (MysticGarden project)
- if we add kotlinx-serialization as dependency, how much bigger does the library become? Currently it is very small because it has no dependencies at all besides standard kotlin. I'd like to keep it that way and ideally don't require the strong dependency at all.
- Can we replace the entity value class with a typealias? What downsides does that have? Does normal `Int` then also get the entity extensions? If so, I don't like that solution because it polutes standard classes. 
- Entity as an interface: what impact does that have to performance?
- Can we "generalize" the Fleks world and allow the user to override the used entity class and entity service? Per default it will use the current implementation. 
  - Maybe something like:
  ```kotlin
  val world = world {
    entityService = useDefault<UserCustomEntity>() // <-- default entity service with custom entity class
    // OR
    entityService = UserEntityService() // <-- custom service by the user with either his own entity class or the Fleks entity
  }

  // any custom entity class must provide an 'id' property to still work with the component index
  // any custom entity service must provide the current entity service functions like create/remove/...
  ```
  - not sure how easy that is and if it is even possible because entity configuration and therefore also component configuration is happening in the current entity service. Not sure how a user could still provide this functionality in his own service.
    - Maybe don't allow the complete service "replacement". Just allow the ID creation/removal and keep all the other things like component assignment, etc. in the Fleks service. So just a part of the current Fleks entity service can be customized by users.
  - However, this approach might be good for some other issues that were discussed lately where people needed their own way to create/assign IDs.

jobe-m:
> is kotlinx-serialization the standard to serialize/deserialize stuff?

kotlinx.serialization is the standard for serializing in Kotlin multiplatform. It works for jvm, js/web and native targets. Downside is that it needs `@Serializable` annotation on each non-primitive object which should  be serialized. So that the Kotlin compiler plugin can automatically create code needed for serialization of the objects.

[...]

PR #92, opened 2023-05-22
add kotlin serialization dependency

Quillraven:
As discussed in #87, kotlinx serialization is the standard for multiplatform serialization. I checked the size difference of built maven files and seems like it is at most 3kb (JVM). For other platforms it is even smaller. That's why I don't see an issue in adding this dependency.

This is the newest version (1.5.1) which required a slightly different gradle setup than the previous 1.4.0 version. Can you try this version please @jobe-m and let me know if it is still working for you?

Also, this is just part 1 for #87. I still want to provide an option to users to use their own "entity id creation/deletion" provider, if necessary. This will be the second part which I also want to include in upcoming 2.4 version.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,6 +4,7 @@ import kotlinx.benchmark.gradle.KotlinJvmBenchmarkTarget
 
 plugins {
     kotlin("multiplatform") version "1.8.21"
+    kotlin("plugin.serialization") version "1.8.21"
     id("org.jetbrains.kotlinx.benchmark") version "0.4.8"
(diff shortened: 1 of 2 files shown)


jobe-m:
This should work with Korge-fleks without problems since Korge is defining the used kotlinx.serialisation version via kproject config during compilation of the Korge project. But will try it out also asap.

jobe-m:
I quickly configured the commit of this pull request in kproject config of Korge-fleks and tested it. Runs without problems 👍 
Feel free to merge :)

PR #93, opened 2023-05-25
add new 'EntityProvider' interface and configuration

Quillraven:
This is a "quick&dirt" implementation of extracting the entity creation/removal of the `EntityService`. That change allows users to provide their own creation logic and adds additional flexibility (part 2 of #87).

Per default, it is still using the normal Fleks entity recycling implementation.

I created a new interface called `EntityProvider` (up for discussion):
```kotlin
interface EntityProvider {
    fun numEntities(): Int

    fun create(): Entity

    fun create(id: Int): Entity

    operator fun minusAssign(entity: Entity)

    operator fun contains(entity: Entity): Boolean

    fun reset()

    fun forEach(action: World.(Entity) -> Unit)
}
```

You can pass your own provider in the configuration with the new `entityProvider` function which calls factory method that takes a `world` instance.

I am personally not sure if this helps the guys who had some questions regarding network games. I never needed such a thing in my own games but that doesn't mean that it can't be useful to others ;)

Tests and comments are currently missing. I just want to get some feedback, if we should add this change or if we leave it as it is. What do you guys think?
@jobe-m @Gidroshvandel

### DIFF
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/world.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/world.kt
@@ -200,6 +200,10 @@ class WorldConfiguration(@PublishedApi internal val world: World) {
         world.setEntityRemoveHook(hook)
     }
 
+    /**
+     * Sets the [EntityProvider] for the [EntityService] by calling the [factory] function
+     * within the context of a [World]. Per default the [DefaultEntityProvider] is used.
+     */
     fun entityProvider(factory: World.() -> EntityProvider) {
         world.entityService.entityProvider = world.run(factory)
     }
@@ -250,7 +254,7 @@ class World internal constructor(
         private set
 
     @PublishedApi
-    internal var entityService = EntityService(this, entityCapacity)
+    internal val entityService = EntityService(this, entityCapacity)
 
     /**
      * List of all [families][Family] of the world that are created either via
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/world.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/world.kt
@@ -55,8 +55,6 @@ class InjectableConfiguration(private val world: World) {
 class ComponentConfiguration(
     @PublishedApi
     internal val world: World,
-    @PublishedApi
-    internal val systems: List<IntervalSystem>,
 ) {
 
     /**
@@ -67,10 +65,6 @@ class ComponentConfiguration(
         type: ComponentType<T>,
         noinline hook: ComponentHook<T>
     ) {
-        if (systems.isNotEmpty()) {
-            throw FleksWrongConfigurationOrderException()
-        }
-
         world.setComponentAddHook(type, hook)
     }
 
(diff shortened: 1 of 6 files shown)


jobe-m:
The implementation looks ok for me. Since the default implementation does not differ from the current implementation I thing we can go for it.

Regarding the multiplayer use case where it is needed to send entities to other clients: By implementing an own Entity provider it should now be possible to add a entity-ID sync mechanism between all game clients. I mean in the create function it would be possible to limit the IDs for a game client to a specific number range which would be unique for each client. The entity ID range for each game client could be managed by a game server (or the main game client). So when sending entities to other clients than an entity ID clash should not happen. Of course it could be a bit more complicated to implement this but now it should be possible at all without changing Fleks code.

src/commonMain/kotlin/com/github/quillraven/fleks/world.kt
```diff
@@ -199,6 +199,10 @@ class WorldConfiguration(@PublishedApi internal val world: World) {
     inline fun onRemoveEntity(noinline hook: EntityHook) {
         world.setEntityRemoveHook(hook)
     }
+
+    fun entityProvider(factory: World.() -> EntityProvider) {
+        world.entityService.entityProvider = world.run(factory)
```
Quillraven, line 204: Need to add a safety here that this gets called first before any other configuration. Because I think otherwise we have a problem when systems are creating entities in their constructor then they will use the Fleks EntityProvider instead of the custom one.

Need to verify that in a test case. Fixing it is easy since we do already something similar with other configuration parts.

Quillraven: I think that we can improve this in general by not immediately calling `run` for all configurations like:

```kotlin
fun injectables(cfg: InjectableConfiguration.() -> Unit) = injectableCfg.run(cfg)

fun components(cfg: ComponentConfiguration.() -> Unit) = compCfg.run(cfg)

fun families(cfg: FamilyConfiguration.() -> Unit) = familyCfg.run(cfg)

fun systems(cfg: SystemConfiguration.() -> Unit) = systemCfg.run(cfg)
```

If we delay the run calls, then the user should be able to specify the things in any order he likes and it should still work. We need to adapt `configureWorld` to something like:

```kotlin
val worldCfg = WorldConfiguration(newWorld).apply(cfg)

// NEW STUFF HERE (call configurations in correct order)
worldCfg.injectablesCfg() // = worldCfg.injectablesCfg.run(cfg)
worldCfg.entityProviderFactory()
worldCfg.systemsCfg()
worldCfg.familyCfg()
worldCfg.componentsCfg()

// assign world systems afterward to resize the systems array only once to the correct size
newWorld.systems = worldCfg.systems.toTypedArray()
```

Quillraven:
I think I was able to refactor the `WorldConfiguration` in a way, so that it does not matter to the user anymore which blocks he defines first.

Before, it was important to have the `systems` block last, because you can e.g. create entities in a system's constructor and therefore also the different hooks must be already registered to trigger them correctly.

I now simply store the configuration lambda and run it afterwards in the order that we need it. All tests are still working so I am confident that this will not break anything.

----

@jobe-m : If you have time, can you please build this branch and try it out in some of your KorGE examples if everything is still working?

Quillraven:
I merge the PR now and will make another push for some version updates (Kotlin, Dokka, ...). Master is then the RC for version 2.4. I will find some time this weekend to test it in my different games.

Hope you can also find some time in the near future with some KorGE examples @jobe-m. But don't stress yourself with it ;) It is not a problem if you don't have time.

jobe-m:
No problem. I am anyway using locally always the latest commit in Korge-fleks. So quick testing should be a bliss.

PR #100, opened 2023-07-03
minor README tidyings

aSemy:
* Applied some IJ quick fix grammar/spelling suggestions
* fixed code styling for XML block
* added `-jvm` suffix for Maven dependency (which is required for non-Gradle projects)
* formatted the benchmark result table


Quillraven:
Thanks for the fixes! Did not know that Maven needs `-jvm` but good to know for the future.

Issue #103, opened 2023-07-06
Proposal for component lifecycle methods


geist-2501:
Hi all!

I've been tinkering with Fleks and, while the component hooks are lovely and work perfectly, I wondered why there isn't lifecycle methods on components instead. 

My specific use-case is that my components can raise events in order to talk to the UI. To do this they need an event manager which I inject - however, components don't have the whole Fleks context thing so I can't inject directly. So instead, I have to initialise it via a hook (which _do_ have a Fleks context), which works great, but took me a while to think of 🗿.

Wouldn't lifecycle methods be more developer friendly? For instance;

```kotlin
class ToolbarComponent : Component<ToolbarComponent> {

    private lateinit var eventManager: EventManager
    
    override fun World.onAddComponent() {
        eventManager = inject()
    }

    // Instead of;
    
    companion object : ComponentType<ToolbarComponent>() {
        val onAddComponent: ComponentHook<ToolbarComponent> = { entity, toolbarComponent ->
            toolbarComponent.eventManager = inject()
        }
    }
}
```

I've forked this repo and got the lifecycle methods working, but I'd be interested to know peoples opinions on whether this is actually a better approach :)

Also Fleks is awesome! Very happy that I made the switch from Ashley halfway through my project.

Quillraven:
My reason for it was that I thought you don't need these methods for every single component AND performance reasons.

What would be interesting for me is how much slower it gets when we add onAdd/onRemove as default functions to a component and call it in the specific ComponentHolder functions where the hooks are currently called.

There is a benchmark sourceSet with a manual.kt file in it. You can run it before and after your changes for a quick comparison. Run it a few times and take the average values.

This is what I usually do to test such changes. The gradle benchmarkJvm task is the real one but takes veeeery long, that's why I don't recommend it for quick testing 😉

Also, ignore the result in the comments of manual.kt file. You will most likely have different numbers. Those are the ones from my PC which was also used for the benchmark table in the README.md. 

My laptop e.g. has completely different values and you will most likely also have different ones. That's why you should compare it with your own numbers and run it before and after your changes.

geist-2501:
Ah! I never considered that it could have a performance impact - I figured it would be the same as with the hooks.

[...]

PR #104, opened 2023-07-10
Component lifecycle methods

geist-2501:
This PR adds component lifecycle methods `onAddComponent()` and `onRemoveComponent`, which replaces the component hooks. As such, this PR also removes component hooks.

Discussion of this change can be found in issue #103.

The TL;DR of #103 is that lifecycle methods have are just a better developer experience over component hooks, with next to no performance hit.

Tests and code documentation have also been updated.

The removal of component hooks has resulted in _slightly_ uglier tests, and I've pointed these out.

### DIFF
--- a/src/commonTest/kotlin/com/github/quillraven/fleks/Fleks2TDD.kt
+++ b/src/commonTest/kotlin/com/github/quillraven/fleks/Fleks2TDD.kt
@@ -20,6 +20,29 @@ private data class Position(
     companion object : ComponentType<Position>()
 }
 
+abstract class ColliderService {
+    abstract fun getId(): String
+}
+
+private data class Collider(
+    val size: Float
+) : Component<Collider> {
+    override fun type(): ComponentType<Collider> = Collider
+
+    var colliderId: String? = null
+
+    override fun World.onAddComponent() {
+        val provider = inject<ColliderService>()
+        colliderId = provider.getId()
+    }
+
+    override fun World.onRemoveComponent() {
+        colliderId = null
+    }
+
+    companion object : ComponentType<Collider>()
+}
+
 private data class Sprite(
     val background: Boolean,
     var path: String = "",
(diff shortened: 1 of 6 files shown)


src/commonTest/kotlin/com/github/quillraven/fleks/Fleks2TDD.kt
```diff
@@ -20,6 +20,29 @@ private data class Position(
...
+        val provider = inject<ColliderService>()
+        colliderId = provider.getId()
+    }
+
+    override fun World.onRemoveComponent() {
+        colliderId = null
+    }
+
+    companion object : ComponentType<Collider>()
+}
```
geist-2501, line 44: Since component hooks could be defined per-test and component lifecycle methods can't, I figured a separate component would be appropriate to test the specific behaviour instead of retrofitting existing ones. I also included an injectable to make it more of an E2E use-case for the lifecycle methods.

Quillraven: Good idea - thanks!

src/commonTest/kotlin/com/github/quillraven/fleks/WorldTest.kt
```diff
@@ -50,9 +58,13 @@ private class WorldTestIteratingSystem(
...
-private class WorldTestInitSystem : IteratingSystem(family { all(WorldTestComponent) }) {
+private class WorldTestInitSystem(
+    onCreateComponent: (WorldTestComponent) -> Unit = {}
+) : IteratingSystem(family { all(WorldTestComponent) }) {
     init {
-        world.entity { it += WorldTestComponent() }
+        val comp = WorldTestComponent()
+        onCreateComponent(comp)
+        world.entity { it += comp }
     }
```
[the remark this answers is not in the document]
Quillraven: I also don't like that structure. I think the test just creates a single entity in the constructor of this system, correct? It should be then possible to simply get the first entity of the world and its WorldTestComponent.

Please try that and remove the `onCreateComponent`. I also find it confusing to read or understand its purpose when just quickly looking at it. Getting the entity and component after the world creation with the usual getters should be feasible and easier to read and understand, I hope.

Quillraven (changes requested):

Quillraven:
Thank you for your contribution! I have one minor remark about the test with the callback as system constructor argument. Should be possible to restructure the tests without using that.

Please have a look and then we are ready to merge 😁

geist-2501:
Replaced it with a much nicer getter in the test itself 👍

Quillraven (approved):

Quillraven:
Cool, great stuff! I will make the SNAPSHOT release on maven tomorrow. And with that I will close the 2.4 features as well. I will do some testing on the weekend and if nothing crazy pops up, then we are good to go for the next stable release!

Thank you again for the support 🙏

PR #105, opened 2023-07-13
Added entities to the arguments for lifecycle methods

geist-2501:
I was updating the Fleks Wiki to remove component hooks in place of the new lifecycle methods, when I realised the example in the Wiki was no longer possible 🗿.

```kotlin
val onAdd: ComponentHook<Box2dComponent> = { entity, b2dCmp ->
    b2dCmp.body = inject<PhysicWorld>().createBody( /* body creation code omitted */)
    b2dCmp.body.userData = entity // No longer possible, no entity reference from the component!
}
```

So instead we should pass the respective entity as a parameter to the component lifecycle methods.

```kotlin
override fun World.onAddComponent(entity: Entity) {
    body = inject<PhysicWorld>().createBody( /* body creation code omitted */)
    body.userData = entity // 🎉
}
```

### DIFF
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/component.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/component.kt
@@ -51,12 +51,12 @@ interface Component<T> {
     /**
      * Lifecycle method that gets called whenever a [component][Component] gets set for an [entity][Entity].
      */
-    fun World.onAddComponent() = Unit
+    fun World.onAddComponent(entity: Entity) = Unit
 
     /**
      * Lifecycle method that gets called whenever a [component][Component] gets removed from an [entity][Entity].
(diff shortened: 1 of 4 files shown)


Quillraven (changes requested):
I reserved some time for Saturday to update my example projects but it is good that you discovered that before!

However, it is bad that this wasn't captured by the tests. Could you please modify one of the tests or create a new one to verify that the correct entity gets passed to the lifecycle methods and also that you can correctly access the component inside the methods? I obviously missed that case 😅

geist-2501:
Yeah it was my bad, I'm 100% certain the old tests for component hooks did have an assert for the expected entity but I removed it and never replaced it 😅.

Quillraven (approved):

PR #107, opened 2023-07-17
tweak entity modification KDocs

aSemy:
Hi 👋, just a small suggestion for docs about the entity modification restriction, to make them more clear. This is my subjective suggestion, so feedback would be very welcome!

I thought I'd make a PR directly so it's easier to view my suggested changes, I hope that's okay!

### Summary

- fix typo, `somOtherEntity` -> `someOtherEntity`
- add correct example to contrast with incorrect usage
- use emoji to highlight correct/incorrect approaches

### DIFF
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
@@ -64,10 +64,14 @@ abstract class EntityComponentContext(
      * **Attention** Make sure that you only modify the entity of the current scope.
      * Otherwise, you will get wrong behavior for families. E.g. don't do this:
      *
-     *     entity.configure {
-     *         // don't do this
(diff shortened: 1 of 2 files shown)


aSemy:
Actually, thinking about it, could this conflict be prevented by using a `@DslMarker`? If there was a `@EntityModificationScope` DSL marker added to `EntityUpdateContext` and `EntityCreateContext`, could that prevent such accidental updates?

Quillraven:
I have read the documentation about @DslMarker again yesterday and imo it cannot help here. The marker helps that you don't nest DSL functions in an unintended way. But we actually want to achieve the arguments passed to the DSL function are restricted in a special way.

I don't think that this is feasible with a DslMarker. Still, if you have a good idea to solve that without impacting performance to much, please feel free to create a new PR for it.

Thanks for the documentation fixes!

PR #108, opened 2023-07-17
Adjust auto-publishing workflow

aSemy:
Adjust the 'publish' workflow so that it's called by the 'build' workflow as a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows)

This hopefully fixes the issue described in #106 

> @aSemy : any idea why "Publish" workflow gets triggered for PRs? I thought the if condition in the job should prevent that? Obviously, I am wrong :D
> 
> _Originally posted by @Quillraven in https://github.com/Quillraven/Fleks/issues/106#issuecomment-1637548926_

I'm not 100% certain that this will solve the issue, so please double check it!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,7 @@
 name: Build
 on:
+  release:
+    types: [ created ]
   push:
     branches:
       - master
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,3 +55,9 @@ jobs:
         with:
           gradle-home-cache-cleanup: true
           arguments: build --stacktrace
+
+  publish:
+    # only publish after 'build' was successful, and this workflow is running on master branch
+    needs: build
+    if: github.ref == 'refs/heads/master'
+    uses: ./.github/workflows/publish.yml
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,5 @@
 name: Build
 on:
-  release:
-    types: [ created ]
   push:
     branches:
       - master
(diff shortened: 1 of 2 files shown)


aSemy:
Seems to work as expected for pull requests, so the publish workflow is skipped

image


Hopefully the publish workflow will be triggered correctly when 'build' runs on the master branch!

aSemy:
I see that the 'publish' workflow is still triggered: https://github.com/Quillraven/Fleks/actions/runs/5574312973/workflow

This is because GitHub is still picking up the unmodified `publish.workflow` that's on the master branch. The `at 1336150` links to the master branch version: https://github.com/Quillraven/Fleks/commit/1336150553e685db69ac2b7b35057f245a566ed6

image

.github/workflows/build.yml
```diff
@@ -1,5 +1,7 @@
 name: Build
 on:
+  release:
```
Quillraven, line 3: Please remove this `release` part. Imo it is not necessary since publishing happens automatically now and if something goes wrong we have the manual `workflow_dispatch` if needed. 

Imo triggering it also on release creation will trigger to much publishing/build actions.

Quillraven (changes requested):
Let's merge it after the small change (if you agree) and then test it ;)

Thank you for your quick look into this issue.

.github/workflows/build.yml
```diff
@@ -1,5 +1,7 @@
...
+    types: [ created ]
```
aSemy, line 4:
```suggestion
```

.github/workflows/build.yml
aSemy: makes sense, done!

Quillraven (approved):

PR #111, opened 2023-08-09
Optimization and cleanup for bit arrays.

tommyettinger:
This is organized logically by commit, and each commit has a (sometimes long-winded) comment. Long story short, addRemove() in the benchmarks goes from 1327 ops/second to 2161 ops/second. The number seems large because I tested with 1000 entities (instead of 10,000) and 5 world updates (instead of 1000) so that I could get meaningful results with short-ish iteration times. That mattered because the benchmark seems to have problems running out of heap if it spends too long on one iteration, though I don't think I encountered this with Fleks -- it was an issue with the Artemis benchmark.

There's lots of bitwise magic here. I'm more than happy to explain any part that doesn't make sense. Bitwise operations are usually quite fast compared to anything that deals with objects, so I'm always happy to see them being put to productive use. The main thing I tried to do was to strip down any of the places where indices 0-63 were looped over in each Long, since that takes more time than necessary in a sparse Long (one with few bits set). There's a new clearEnsuringCapacity() method in MutableEntityBag; it is functionally just like:

```kt
clear()
ensureCapacity(capacity)
```

But avoids copying the empty array without the need to. This is something that some of libGDX's collections also have.

Let me know what I can do to help!

### DIFF
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/collection/bitArray.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/collection/bitArray.kt
@@ -43,7 +43,12 @@ class BitArray(
     fun set(idx: Int) {
         val word = idx ushr 6
         if (word >= bits.size) {
-            bits = bits.copyOf(max(word + word, 1))
+            // magic.
+            // ((word or -word) shr 63) is an idiom that returns 0 if word is 0, or -1 otherwise.
+            // Calling inv() on that makes it -1 if word is 0, or 0 otherwise.
+            // With the rest, this makes the copy use twice as many Long items as before, or use
+            // one Long item if word is 0 (and bits.size is currently 0, a precondition).
+            bits = bits.copyOf(word + word - ((word or -word) shr 63).inv())
         }
         bits[word] = bits[word] or (1L shl idx)
     }
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/collection/bitArray.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/collection/bitArray.kt
@@ -17,10 +17,10 @@ class BitArray(
     nBits: Int = 0
 ) {
     @PublishedApi
-    internal var bits = LongArray((nBits + 63) / 64)
+    internal var bits = LongArray((nBits + 63) ushr 6)
 
     val capacity: Int
-        get() = bits.size * 64
+        get() = bits.size shl 6
 
     val isNotEmpty: Boolean
         get() {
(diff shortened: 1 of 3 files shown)


src/commonMain/kotlin/com/github/quillraven/fleks/collection/bitArray.kt
```diff
@@ -36,30 +36,31 @@ class BitArray(
...
+        val word = idx ushr 6
+        return word < bits.size && (bits[word] and (1L shl idx)) != 0L
     }
 
     fun set(idx: Int) {
-        val word = idx / 64
+        val word = idx ushr 6
         if (word >= bits.size) {
-            bits = bits.copyOf(word + 1)
+            // magic.
```
Quillraven, line 46: I have some questions here:
- why is this approach better (and faster?) then a simple `word + 1`?
- Your approach always created a bits array with a size that is a multiple of 2. I could imagine this has a positive effect as well since normally multiple of 2 things are faster?


Otherwise, I suggest to remove `magic.` and the `With the rest....` comment since your other comment explains it very well and at least I don't care about the comparison of a previous approach. We anyway have a git history, if someone is interested ;)

Quillraven:
Very cool stuff, thank you! I thought I removed the redundant `% 64` operations in `BitArray` but seems like I forgot to push at some point and it got lost. You fixed that now as well :)

Just fyi, I also get the heap memory issue in the Artemis benchmark from time to time. Normally a simple restart of the benchmark solves it.

src/commonMain/kotlin/com/github/quillraven/fleks/collection/bitArray.kt
tommyettinger: Part 1: So the reason `word + 1` might be slower is that it only resizes a little bit at a time -- so if you start at size 1 and gradually set() larger values, it has to allocate a new bits Array every time you set a larger value. This can be an issue for speed initially, though not a big one, but it's a bigger problem for garbage collection, since this could create quite a few Array items in one frame, all but one of which need GC-ing. By doubling the array size, the total number of resizes possible for an Array maxes out at 31 or less (the maximum size of an Array on the JVM is just shy of 2 to the 31, about 2.1 billion); when it is already setting inside the maximum size, no resize is needed. It does use more memory as a maximum for data that isn't being modified, but it discards far fewer Arrays along the way. If you know your data only grows to small maximum sizes, then adding one to the size isn't that bad, and it ensures optimal memory usage when not setting any bits. This is a pretty classic tradeoff between best-case memory usage and best-case (and average-case) speed. To me, the stuttering associated with garbage collection makes me really try to avoid it in game code.

Part 2: I know it isn't clear code; that's why I tried to comment it. I wasn't comparing the previous approach at all; I was describing the resize logic in terms of before resize and after resize.

Addendum: It may be worth benchmarking if this equivalent code is faster:

`bits = bits.copyOf(max(word + word, 1))`

The performance of max() varies across JVM versions, but the bitwise code used now should be stable. The bitwise code is also branchless, which doesn't have as much effect as one would think due to speculative execution and such. I'm not sure if there's some compiler code that makes max() branchless or effectively so.

Issue #113, opened 2023-09-13
Safe way to reference entities?


dragbone:
I'm working on a game prototype using fleks and I'm struggling to find a good way to reference an entity from a component. A simple example would be one entity following another entity and therefore needing a reference. 
It seems that the Entity value class is not a safe reference because entities get reused after removal from the world, so a reference might reference a completely different entity in the future.
Is there some safe way to reference entities? Do I need to set a remove entity hook and search for all components that might reference that entity?

Quillraven:
Hi,

this is a good question and tbh I don't know what are the best practices here.

I remember that I did what you mention at the end in an old project using Ashley and I didn't like it because every time I add a component with entity references I had to remember to add the cleanup logic to the hook. Also performance wise it is not ideal because you have to iterate through all such components and also through the entity ref containers.

Another solution which is maybe better and which I also used in some projects is a RemoveSystem with a RemoveComponent. Instead of removing an entity directly you then have to add a RemoveComponent to it. The system then really removes the entity.
This has the advantage that you can control when entities get really removed. It is also easy to e.g. add a timer to en entity and remove it delayed by simply adding a delay to the RemoveComponent and adjust the logic in the RemoveSystem.

Also, on your entity ref components you can then check if your entity has a RemoveComponent and remove it or do whatever logic you need to do (e.g. follow another nearby entity).


Finally, in the last version of Fleks you can provide your own EntityProvider. Check the wiki. That way you can remove the recycle logic and always create new IDs. But be careful to not overflow the IDs.


Of course I am open for other ideas and can add some generic solution to Fleks if there is a good one 😃

dragbone:
> Finally, in the last version of Fleks you can provide your own EntityProvider. Check the wiki. That way you can remove the recycle logic and always create new IDs. But be careful to not overflow the IDs.

That actually sparked an idea! What I want is to identify different versions of the same entity, so I hacked together some utilities to actually track versions of entities: https://gist.github.com/dragbone/149b63c55eec650a32bf60d697b80fc0

A generic solution could probably work pretty similar by tracking versions internally and providing references when needed (without attaching a Component to every entity).

Quillraven:
Is this a convenient enough solution for you? Looks pretty minimalistic and nice.

Out of curisoity: would it help if recycled entities don't get reused within a single frame (=world.update call)? I think that should also work at least if the important systems are triggered every frame. If you have systems that run at a fixed rate then this won't help. In that scenario your solution is better and should still work.

[...]

PR #114, opened 2023-09-15
#113 entity version

Quillraven:
PR for #113 (WIP).

@dragbone: I quickly run the benchmarks when changing `Entity` from a value class to a data class.

```kotlin
// from this
@JvmInline
@Serializable
value class Entity(val id: Int)


// to this
@Serializable
data class Entity(val id: Int, val version: Int = 0)
```

To my surprise, this does not seem to have any impact on the performance. Maybe because everything internally still works with the id (=int) and somehow that already gets optimized to perfection by the compiler/JVM? Don't know about the otheer platforms.

Here is the result:
```
# value class
jvmBenchmarks summary:
FleksBenchmark.addRemove  thrpt    5  3005,834 ± 104,474  ops/s
FleksBenchmark.complex    thrpt    5     1,515 ±   0,308  ops/s
FleksBenchmark.simple     thrpt    5    91,181 ±   5,372  ops/s

# data class
jvmBenchmarks summary:
FleksBenchmark.addRemove  thrpt    5  3011,912 ± 216,568  ops/s
FleksBenchmark.complex    thrpt    5     1,788 ±   0,094  ops/s
FleksBenchmark.simple     thrpt    5    87,405 ±  32,696  ops/s
```

Imo we can go ahead with this change as it is a nice feature when it comes to safe referencing entities in components and it seems to have no negative effect from what I can tell.

----

One thing that comes to my mind is serialization. Before it was just the id of the entity but now you get an id + version. The version is imo irrelevant for seralization because when you load a save state you most likely start again with version 0 for all entities?

@jobe-m: do you have any concerns about changing `Entity` from a value class to a data class?

@dragbone: If you like, you can start from this branch and add the logic to increase the version when an entity gets reused and add the tests for it.
The correct place should be the `create` method of the `DefaultEntityProvider`. And also the `contains` method of the `DefaultEntityProvider` to check for the correct version.

If you don't have time then I will try to finalize it in October.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -49,6 +49,11 @@ kotlin {
 
 benchmark {
     configurations {
+        create("FleksOnly") {
+            include("Fleks")
+            exclude("Artemis|Ashley")
+        }
+
         create("FleksAddRemoveOnly") {
             include("addRemove")
             exclude("Artemis|Ashley")
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/collection/bag.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/collection/bag.kt
@@ -42,6 +42,10 @@ class Bag<T>(
         return values[index] ?: throw NoSuchElementException("Bag has no value at index $index")
     }
 
+    inline fun getOrNull(index: Int): T? {
+        return values[index]
+    }
+
     fun hasNoValueAtIndex(index: Int): Boolean {
         return index >= size || values[index] == null
     }
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/collection/bag.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/collection/bag.kt
@@ -61,6 +61,15 @@ class Bag<T>(
         return false
     }
 
+    fun removeIndex(index: Int): Boolean {
+        if (values[index] != null) {
+            values[index] = values[--size]
+            values[size] = null
+            return true
+        }
+        return false
+    }
+
     fun clear() {
         values.fill(null)
         size = 0
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
@@ -11,6 +11,8 @@ import kotlin.jvm.JvmInline
 @Serializable
 value class Entity(val id: Int)
 
+data class EntityRef(val entity: Entity, var valid: Boolean = true)
+
 /**
  * Type alias for an optional hook function for an [EntityService].
  * Such a function runs within a [World] and takes the [Entity] as an argument.
@@ -81,6 +83,8 @@ abstract class EntityComponentContext(
      * future calls to [World.entity].
      */
     fun Entity.remove() = componentService.world.minusAssign(this)
+
+    fun Entity.ref() = componentService.world.entityService.entityRef(this)
 }
 
 /**
@@ -375,6 +379,9 @@ class EntityService(
     internal val compMasks = bag<BitArray>(initialEntityCapacity)
 
     @PublishedApi
+    internal val entityRefs = bag<EntityRef>(initialEntityCapacity)
+
+    @PublishedApi
     internal val createCtx = EntityCreateContext(compService, compMasks)
 
     @PublishedApi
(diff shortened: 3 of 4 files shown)


dragbone:
@Quillraven I think the version has to be serialized as well or else any references to entities might break due to versions no longer matching.
I  have looked into how fleks works internally a bit and the change looks like it might be more complicated than expected. There are many places where the Entity value class gets created from just an index (e.g. Family.foreach creates Entity references from its BitArray), which now would need access to the current entity version to create a valid reference.
I think these lookups will have a performance impact, how much is hard to say without implementing it...

Quillraven:
Ah okay, that explains why my change did not have any bigger impact. So we need to test that again, when all areas that create an Entity like `Entity(someindex)` are adjusted as well.

About the serialization: I still think it is not necessary because what is the use-case here? Usually you want to save a state of the world and then load it later on. When you load it you start from an empty world or "override" the current state of the world meaning that all entities are new and should start with version 0 again.

Why would we need the version?

dragbone:
> About the serialization: I still think it is not necessary because what is the use-case here? Usually you want to save a state of the world and then load it later on. When you load it you start from an empty world or "override" the current state of the world meaning that all entities are new and should start with version 0 again.
> 
> Why would we need the version?

I image users expect that creating a snapshot and restoring it would result in a world that is exactly the same as it was before. That includes that all entity references that might be stored in some Components are still valid after the restore and that is only possible if all entities have their version restored.
I haven't used the snapshot feature yet so that might be a wrong assumption on my part.

Quillraven:
@dragbone: I quickly drafted the idea of the `handle` that I mentioned in your PR. I called it `EntityRef` instead because I liked that name better.

In short:
- There is a new `ref()` extension function for an `Entity` that returns an `EntityRef`
- This reference is `valid` until the entity gets removed

Also added a test to quickly verify the behavior. A component needs to use `EntityRef` instead of `Entity` when it needs a safe reference:
```kotlin
data class EntityRefComponent(val ref: EntityRef) : Component<EntityRefComponent> {
    override fun type() = EntityRefComponent

    companion object : ComponentType<EntityRefComponent>()
}
```

What do you think about this approach? Which one do you like more? The versioning or the ref? Which one feels better in your specific game?

dragbone:
From a usability point of view I'd prefer directly getting entity references that i can store and use without having to think about it. Having to remember to call `ref()` is somewhat error prone and not very self-explanatory.
But in the end I can live with both ¯\\\_(ツ)\_/¯

Quillraven:
@dragbone:

I fiddled around a little bit now an this would be a working entity value class:

```kotlin
/**
 * An entity of a [world][World]. It represents a unique identifier that contains several infos:
 * - index of the entity (first 32 bits)
 * - version of the entity (next 16 bits)
 * - optional flags (last 16 bits)
 */
@JvmInline
@Serializable
value class Entity(val id: Long) {

    // To explain the bitwise operations:
    // To extract n bits of a long from a specific offset we can use following approach:
    //
    // val rightShifted = long >>> offset
    // val mask = (1L << nBits) - 1L
    // val result = rightShifted and mask
    //
    // 32bits mask can be represented as 0xFF FF FF FF
    // 16bits mask can be represented as 0xFF FF

    fun nextVersion(): Entity {
        val v: Long = version + 1L
        val newId = id and 0xFFFF0000FFFFFF
        return Entity(newId or (v shl 32))
    }

    val idx: Int
        get() = (id and 0xFFFFFFFF).toInt()

    val version: Int
        get() = ((id ushr 32) and 0xFFFF).toInt()

    companion object {
        val NONE = Entity(-1)
    }
}
```

HOWEVER, I now understand better what you meant with the different lookups and imo it makes the codebase harder to understand and at least in my branch the addRemove benchmark performance went down quite a lot. Simple and complex were the same.
Also, the `EntityProvider` needs the version function, like you have it in your PR as well, which I also don't like because imo users should not be bothered with such details and it will be most likely tricky to get that right when implementing their own `EntityProvider`.

Therefore, I change my mind and would vote for the `ref` solution for a quick-win solution which is easier to implement, does not affect performance and also does not "polute" the codebase that much.

Fine for you? Or do you want to investigate the value class approach some more?

dragbone:
I had one last idea last night which I want to test maybe today if I have time or else tomorrow.

Quillraven:
#115 has a more convenient solution for the entity version. We will go for that approach. I close this PR draft.

PR #115, opened 2023-09-16
#113 entity version

dragbone:
@Quillraven here is my attempt at implementing issue #113. I'm not entirely happy with the solution... I'm open to any suggestion.

## Notable changes
* EntityProvider got a new function `fun getCurrentVersion(id: Int): Entity` to retrieve the current version of an entity
* I introduced `Entity.NONE` as a shorthand to create an invalid entity (replaces `Entity(-1)`)

## Open tasks
* [x] Javadoc for `EntityProvider.getCurrentVersion(id: Int)`
* [x] `DefaultEntityProvider.minusAssign(entity: Entity)` has to update the entity version lookup for snapshots to work correctly, I think that's an ugly workaround. Snapshot loading seems kinda hacky in general, maybe a refactoring is in order now.

## Benchmarks
Benchmarking this is a bit different because the benchmark now needs to track the references to the entities. I have run the adjusted benchmark on master as well to get a result that can be compared, you can expect 20%-25% slowdown in the addRemove benchmark, the other two benchmarks seem unaffected:

#### master with benchmark adjustment
| Benchmark                 | Mode  | Cnt   |    Score  |    Error  | Units  |
| ------------------------- | ----- | ----- | --------- | --------- | ------ |
| FleksBenchmark.addRemove  | thrpt |    5  |  2636.779 |  ± 37.850 |  ops/s |
| FleksBenchmark.complex    | thrpt |    5  |     1.875 |  ±  0.038 |  ops/s |
| FleksBenchmark.simple     | thrpt |    5  |    90.531 |  ±  8.086 |  ops/s |

#### branch with entity version tracking
| Benchmark                |  Mode | Cnt |    Score |     Error | Units |
| ------------------------ | ----- | --- | -------- | --------- | ----- |
| FleksBenchmark.addRemove | thrpt |   5 | 2093.105 | ± 165.674 | ops/s |
| FleksBenchmark.complex   | thrpt |   5 |    1.965 | ±   0.053 | ops/s |
| FleksBenchmark.simple    | thrpt |   5 |   88.293 | ±  15.189 | ops/s |


Quillraven:
Thanks for the contribution but I will need some time to check it. I hope that I can give an answer/review by Monday.

If only the first benchmark is affected then I think it is not a big deal because thanks to @tommyettinger that got recently improved by a lot and imo adding a clean solution for entity versioning is more important.

----

Just an idea that I got from an article. I think an int has 32 bits in Kotlin for all platforms? Or maybe even 64?

Maybe we can again do some bitwise magic stuff to encode the entity version into the ID as well.

My idea is that e.g. we use the first 16 bits for the version and the last 16 bits for the id itself. Of course that reduces the total amount of possible entities but I think it is still a big number that won't be reached by most games using Fleks.

That way we safe an additional lookup for the version and we can keep the value class and therefore still working with an Int instead of a class object.

I am not 100% sure if that works out but it is just an idea that popped up today in my head. 

What do you think? Could that work out and maybe simplify the current aporoach and keep performance and memory consumption on the same level?

dragbone:
The problem as I see it is not really with how we encode or store the versions. The storing aspect is mostly a non-issue because for usual use cases the memory consumption is dominated by data in components. Performance wise I don't think we gain much by changing the encoding (might even be a loss if we need to decode and encode too often).
The performance impact comes from the fact that e.g. a family only stores which entities are active (in a BitArray) and needs to lookup the current version to use whenever it is iterated. In addition we need to keep that version lookup array updated which also costs some performance.

Quillraven:
@dragbone: Sorry once more! I will have a look tomorrow morning and give you feedback to the PR.

Coming back to the idea with splitting the 32 bit int into two parts for version and id I still think it might be simpler and does not require most of the logic that we came up with right now. But let me elaborate and please let me know, if my thought process is wrong somewhere.

For simplicity reason, let's say we only have a 4 bit number for entities where 0000 is 0 and 1111 = 15. In total 16 possible values.

If we now want to use half for ids and half for versions, we end up supporting only up to 8 entities.

Now imagine following scenario:
- We create an entity (id =0, version=0). The Entity value is 0000.
- In a component we store this value to have a reference to the entity.
- We now remove the entity and by doing so, we increase its version. 0000 -> 0100 (or 0001; I am not sure yet if it is smarter to use the end bits or the front bits for the version)
- We create an entity again. Our removed entity gets recycled. It still has id 0 but its version is 1.
- If we'd check our entity in the component, we'd get the info that it is dead because the number 0000 is not part of our world. We only have 0100.
-- We can e.g. have an additional `EntityBag` in our `EntityProvider` and if the value of the bag at index `Entity.id` is not equal to the entity value passed to the `contains` function then the entity got removed and recycled.
-- In our example it is simply comparing 0000 with 0100 and the index of the bag is 0 (=entity id without version)
- Also, when a family is iterating or an EntityHook gets called then we pass the entity from that additional bag (=id + version) instead of just the id

The only additional thing that is happening is that a separate bag of entities is used and that a bit shift operation is happening when removing entities, which should be not a big deal imo because direct lookup of an array index should not be that expensive.

My idea would be to have something like this, but not sure if that is possible in Kotlin:
```kotlin
// entity is still a value class like before
@JvmInline
@Serializable
value class Entity(val versionId: Int) {
  val id : Int // add id property to make the code compile. However, this should not be a real property because we are in a value class so it is just some syntax sugar in Kotlin
    get() = versionId and and 0xFFFF

  fun incVersion() : Entity  {
    // Mask out the last 16 bits and increment the first 16 bits
    val last16Bits = versionId and 0xFFFF
    val first16Bits = (versionId ushr 16) + 1
    // Combine the first 16 bits with the last 16 bits
    return Entity((first16Bits shl 16) or last16Bits)
  }
}
```

With that change the entire codebase should still compile and all tests should pass because `id` is still available. In addition, we need to increase the version when an entity gets removed and we need to make sure that anywhere we pass the Entity to a user (like in family iteration, EntityHooks, ...) we use the entity value of the new bag.

----

Maybe that is what you did in your implementation already. Again, I will have a look at it tomorrow.
Let me also share the article that I have found: https://skypjack.github.io/2019-05-06-ecs-baf-part-3/
Maybe it provides you with some awesome idea ;)

Quillraven:
Me again! I also chatted with ChatGPT on my phone now a little bit and it gave me two other ideas:
1) have UUIDs per entity
2) use handles for entities

Both strategies don't require the versioning stuff at all.

For solution 1 we'd need some platform specific implementations afaik because there is no build in UUID for kotlin multiplatform.

Solution 2 sounds interesting to me and my idea would be the following:
- When users need safe references to entities in components, they need to work with a handle instead of the `Entity`.
- the handle is a simple container like
    ```kotlin
    data class EntityHandle(val entity:Entity, var valid: Boolean = true)
    ```
- we could provide an extension function to `Entity` similar to the existing `remove` function. E.g. let's call it `handle()`.
- This function returns either the current existing handle of the entity or a new one, if none exists yet.
- We could store this e.g. in a `bag` in the `EntityService` like we do it for the `compMasks`
- When an entity gets removed then we need to set the `valid` flag of its `EntityHandle` to false, if it exists and remove it from the bag.


What I like about this approach is that this new functionality is used "lazily" and does not require the lookups of our previous solution (e.g. in family iterations).

The downside is that users need to be aware to work with `EntityHandle` in such scenarios instead of `Entity`.

What do you think about that approach?

src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
```diff
@@ -2,13 +2,16 @@ package com.github.quillraven.fleks
...
 import com.github.quillraven.fleks.collection.*
 import kotlinx.serialization.Serializable
-import kotlin.jvm.JvmInline
 
 /**
  * An entity of a [world][World]. It represents a unique id.
  */
 @Serializable
-data class Entity(val id: Int, val version: Int = 0)
+data class Entity(val id: Int, val version: Long) {
```
Quillraven, line 10: I think the change from `value class` to `data class` might be bigger than we initially thought. I am really wondering why the benchmarks are not showing that but maybe because they are too short to trigger the garbage collector (GC).

I now had some time to look into more details and there are several places where Fleks is using `Entity(id)` to "create" an entity. For example in `DefaultEntityProvider.create`, `DefaultEntityProvider.forEach`, `BitArray.toEntityBag` or the entire `MutableEntityBag` class.

Granted, you already optimized it by making `Entity(-1)` a constant `Entity.NONE` but in the other areas we are now always creating a new class instance of an entity instead of working with a primitive Int.

I think this is not a good idea and could cause bad behavior especially on mobile devices like Android which are not as fast and efficient when it comes to GC.

----

One way to keep this approach and mitigate this issue, is to keep the `value class` and encode the version into the entity id, but let's see.

src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
```diff
@@ -257,13 +265,16 @@ class DefaultEntityProvider(
      */
     override fun numEntities(): Int = nextId - recycledEntities.size
 
+    private val currentEntityVersions = Bag(Array(initialEntityCapacity) { 0L })
```
Quillraven, line 268: `Int` should be sufficient. Entity is also just an `Int`, therefore it doesn't make sense imo to make the version a `Long`.

src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
```diff
@@ -276,7 +287,7 @@ class DefaultEntityProvider(
                 nextId = recycled.id + 1
             }
 
-            recycled
+            recycled.copy(version = ++currentEntityVersions[recycled.id])
```
Quillraven, line 290: Similar to my comment for the `Entity data class`, we are creating here a new object instance every  time an entity gets recycled. 
I understand why we need to do that because we cannot modify the original instance because otherwise the entity refs in components will not get invalid, but again it creates additional object instances compared to current Fleks.

src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
```diff
@@ -310,12 +322,16 @@ class DefaultEntityProvider(
     override operator fun minusAssign(entity: Entity) {
         recycledEntities.add(entity)
         removedEntities.set(entity.id)
+        currentEntityVersions[entity.id] = entity.version
```
Quillraven, line 325: I think this line is not necessary because the version is only updated in the `create` method before. Assigning it here once more is redundant?

src/commonMain/kotlin/com/github/quillraven/fleks/world.kt
```diff
@@ -477,15 +477,17 @@ class World internal constructor(
             return
         }
 
+        val versionLookup = snapshot.keys.associateBy { it.id }
```
Quillraven, line 480: That's the part of my question from the other PR, correct? I still think that storing and restoring the entity version is not needed for snapshots. But even if it is, we'd need to reset the entity version at some point because otherwise it simply is increasing infinitly and will overflow at some point.

src/commonTest/kotlin/com/github/quillraven/fleks/ComponentTest.kt
```diff
@@ -50,7 +50,7 @@ internal class ComponentTest {
 
     @Test
     fun addEntityToHolderWithSufficientCapacity() {
-        val entity = Entity(0)
+        val entity = Entity(0, version = 0)
```
Quillraven, line 53: I suggest to use 0 as a default value in the entity constructor

dragbone:
@Quillraven I moved some stuff around and replaced some data structures to get rid of the lookup. At least on my machine the performance is now pretty much the same as it is on master.

Short summary of what I did:
* The Entity class now only gets instantiated by EntityProvider (plus one usage in world for snapshot loading - should be removed as well but didn't have time)
* EntityProvider and Family now store a Bag<Entity> of the entities that concern them instead of BitArrays
* Version lookup is no longer needed because the necessary places now have a reference to currently valid entities

I want to write some additional tests to make sure that everything still works, I think there are some gaps in the test coverage that's relevant for this change. I also want to add a benchmark to test if there is a memory/GC issue (I think there isn't, but I'd prefer proof).

src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
```diff
@@ -276,7 +277,9 @@ class DefaultEntityProvider(
                 nextId = recycled.id + 1
             }
 
-            recycled
+            recycled.copy(version = recycled.version + 1)
+        }.also {
+            entities[it.id] = it
```
Quillraven, line 282: That's interesting. I didn't know that you can do that in Kotlin. I learned something new, thanks! :)

src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
Quillraven: Would be great if you can also compare `data class` to `value class` in your memory test. Imo the value class should perform a lot better since it should be just a simple `Int` or `Long` which is more efficient to handle.

Also, if I understood your comment correctly, you anyway were able to pass the entire `Entity` class around without the usage of `BitArray`. If that is the case and if the performance is almost the same then I'd vote to revert to the value class again. My example of the other PR worked fine for me and was doing well until I introduced all the different lookups :D

But let's wait for your findings. I am excited to see the outcome!

src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
```diff
@@ -309,13 +313,15 @@ class DefaultEntityProvider(
...
-        removedEntities.set(entity.id)
+        entities.removeAt(entity.id)
     }
 
     /**
      * Returns true if and only if the given [entity] is active and part of the provider.
      */
-    override fun contains(entity: Entity): Boolean = entity.id in 0 until nextId && !removedEntities[entity.id]
+    override fun contains(entity: Entity): Boolean =
+        entity.id in 0 until nextId &&
```
Quillraven, line 323: I think the first part is not necessary anymore. If the entity is not part of the `entities` bag then it is not part of the `EntityService`.

src/commonMain/kotlin/com/github/quillraven/fleks/entity.kt
```diff
@@ -324,19 +330,17 @@ class DefaultEntityProvider(
...
 
     /**
      * Performs the given [action] for all active [entities][Entity].
      */
     override fun forEach(action: World.(Entity) -> Unit) {
         for (id in 0 until nextId) {
-            val entity = Entity(id)
-            if (removedEntities[entity.id]) {
-                continue
+            entities.getOrNull(id)?.let { entity ->
```
Quillraven, line 341: This can be rewritten to:

```kotlin
entities.forEach{ world.action(entity) }
```

Should be slightly more efficient.

src/commonMain/kotlin/com/github/quillraven/fleks/family.kt
```diff
@@ -108,7 +110,7 @@ data class Family(
             if (isDirty && !isIterating) {
                 // no iteration in process -> update entities if necessary
                 isDirty = false
-                entityBits.toEntityBag(field)
+                privateEntities.toEntityBag(field)
```
Quillraven, line 113: I'd inline the code from below since it is only used once and is only two lines of code

src/commonMain/kotlin/com/github/quillraven/fleks/family.kt
```diff
@@ -163,7 +165,7 @@ data class Family(
     /**
      * Returns true if and only if the given [entity] is part of the family.
      */
-    operator fun contains(entity: Entity): Boolean = entityBits[entity.id]
+    operator fun contains(entity: Entity): Boolean = !privateEntities.hasNoValueAtIndex(entity.id)
```
Quillraven, line 168: Add a new `hasValue` or use `getOrNull(entity.id) != null`. Double negation is always confusing imo :)

src/commonMain/kotlin/com/github/quillraven/fleks/family.kt
```diff
@@ -233,15 +236,17 @@ data class Family(
...
-        if (entityInFamily && !entityBits[entity.id]) {
+        if (entityInFamily && privateEntities.hasNoValueAtIndex(entity.id)) { // !entityBits[entity.id]
             // new entity gets added
             isDirty = true
-            entityBits.set(entity.id)
+            countEntities++
+            privateEntities[entity.id] = entity
             addHook?.invoke(world, entity)
-        } else if (!entityInFamily && entityBits[entity.id]) {
+        } else if (!entityInFamily && !privateEntities.hasNoValueAtIndex(entity.id)) {
```
Quillraven, line 245: refer to comment above about double negation

src/commonMain/kotlin/com/github/quillraven/fleks/family.kt
```diff
@@ -233,15 +236,17 @@ data class Family(
     @PublishedApi
     internal fun onEntityCfgChanged(entity: Entity, compMask: BitArray) {
         val entityInFamily = compMask in this
-        if (entityInFamily && !entityBits[entity.id]) {
+        if (entityInFamily && privateEntities.hasNoValueAtIndex(entity.id)) { // !entityBits[entity.id]
```
Quillraven, line 239: Extracting the `hasNoValueAtIndex` result would be good because in worst case this lookup is executed twice. We can reduce it to one. Not a huge improvement but still ;)

Quillraven:
> @Quillraven I moved some stuff around and replaced some data structures to get rid of the lookup. At least on my machine the performance is now pretty much the same as it is on master.
> 
> Short summary of what I did:
> 
> * The Entity class now only gets instantiated by EntityProvider (plus one usage in world for snapshot loading - should be removed as well but didn't have time)
> * EntityProvider and Family now store a Bag of the entities that concern them instead of BitArrays
> * Version lookup is no longer needed because the necessary places now have a reference to currently valid entities
> 
> I want to write some additional tests to make sure that everything still works, I think there are some gaps in the test coverage that's relevant for this change. I also want to add a benchmark to test if there is a memory/GC issue (I think there isn't, but I'd prefer proof).

That looks neat! If everything is still working and performance is close (or maybe even better? ;)) to what we have before then let's use your PR instead of the `EntityRef` PR. 

Really appreciate your time and effort for this - thank you!

I added some comments as a review and please also share your info, how you do a memory/GC test and how data class compares to value class. Would be cool if we get this solution up and running.

dragbone:
I did some memory tests which didn't show much difference between this version and the one on master. I tested the performance under very harsh memory limitation by setting the maximum heap to 8 megabytes (`@Fork(value = WARMUPS, jvmArgs = ["-Xmx8m"])`) and I did a run with a GC profiler attached by running the benchmark like this:
```
./gradlew jvmBenchmarksBenchmarkJar
java -jar build/benchmarks/jvmBenchmarks/jars/Fleks-jvmBenchmarks-jmh-2.5-SNAPSHOT-JMH.jar -prof gc FleksBenchmark
```
Both didn't indicate much difference between the versions. I'll add some more tests and clean up any comments later this week/on the weekend.

Quillraven:
I also run the benchmarks now on my notebook for master and your fork. I adjusted the versions of your fork before to match master because I updated them a few days ago:

```toml
[versions]

kotlin = "1.9.10"
kotlinxSerialization = "1.6.0"
kotlinDokka = "1.9.0"
kotlinxBenchmark = "0.4.9"
```

Result of master:
```
jvmBenchmarks summary:
FleksBenchmark.addRemove  thrpt    5  3018,381 ± 127,632  ops/s
FleksBenchmark.complex    thrpt    5     1,664 ±   0,062  ops/s
FleksBenchmark.simple     thrpt    5    92,042 ±   7,373  ops/s
```

Result of your fork:
```
jvmBenchmarks summary:
FleksBenchmark.addRemove  thrpt    5  2127,258 ± 521,508  ops/s
FleksBenchmark.complex    thrpt    5     1,654 ±   0,249  ops/s
FleksBenchmark.simple     thrpt    5    94,820 ±   2,222  ops/s
```

Simple and complex seem to have no impact. AdddRemove is slowed down quite a bit BUT it is still super fast and in that case I would vote to introduce the feature and ignore the performance impact in that particular use-case because it is a very nice and important feature.

I also modified your fork and used the `value class` implementation of my PR. This allows us to use `MutableEntityBag` in the `EntityService` and `Family` which simplifies everything down to an `IntArray` basically.
However, with my entity value class implementation the performance was worse. I guess because the index lookup (=id of the entity) is a getter and the bitwise operation needs to be executed every single time when the id gets accessed. 


TL;DR: your implementation looks good and the addRemove performance impact should not be that critical imo. Thank you a lot for your contribution and time investment into this feature!

dragbone:
Remember that the addRemove benchmark is not comparable to master because I had to adjust it to handle entity versions correctly. But I fixed it just now :D https://github.com/Quillraven/Fleks/pull/115/files#diff-e8658b624510082e4daf11283b3a6ac32ee654d77b4eaa8b86fa78ef5034e642
Should now be comparable to master :)

Here are my results from benchmarking with GC profiler:
### master
| Benchmark | Mode | Cnt | Score | Error | Units |
|-|-|-|-|-|-|
| FleksBenchmark.addRemove | thrpt | 25 | 2528.476 ± | 36.232 | ops/s |
| FleksBenchmark.addRemove:·gc.alloc.rate | thrpt | 25 | 1506.890 ± | 21.587 | MB/sec |
| FleksBenchmark.addRemove:·gc.alloc.rate.norm | thrpt | 25 | 729170.015 ± | 1.211 | B/op |
| FleksBenchmark.addRemove:·gc.churn.G1_Eden_Space | thrpt | 25 | 1508.409 ± | 50.651 | MB/sec |
| FleksBenchmark.addRemove:·gc.churn.G1_Eden_Space.norm | thrpt | 25 | 729766.657 ± | 19636.393 | B/op |
| FleksBenchmark.addRemove:·gc.churn.G1_Survivor_Space | thrpt | 25 | 0.045 ± | 0.022 | MB/sec |
| FleksBenchmark.addRemove:·gc.churn.G1_Survivor_Space.norm | thrpt | 25 | 21.518 ± | 10.386 | B/op |
| FleksBenchmark.addRemove:·gc.count | thrpt | 25 | 223.000 | counts |
| FleksBenchmark.addRemove:·gc.time | thrpt | 25 | 134.000 | ms |
| FleksBenchmark.complex | thrpt | 25 | 1.801 ± | 0.107 | ops/s |
| FleksBenchmark.complex:·gc.alloc.rate | thrpt | 25 | 713.654 ± | 41.438 | MB/sec |
| FleksBenchmark.complex:·gc.alloc.rate.norm | thrpt | 25 | 480314367.381 ± | 16654.630 | B/op |
| FleksBenchmark.complex:·gc.churn.G1_Eden_Space | thrpt | 25 | 705.871 ± | 70.504 | MB/sec |
| FleksBenchmark.complex:·gc.churn.G1_Eden_Space.norm | thrpt | 25 | 475086017.877 ± 38615140.377 | B/op |
| FleksBenchmark.complex:·gc.churn.G1_Survivor_Space | thrpt | 25 | 0.030 ± | 0.016 | MB/sec |
| FleksBenchmark.complex:·gc.churn.G1_Survivor_Space.norm | thrpt | 25 | 20318.304 ± | 10186.539 | B/op |
| FleksBenchmark.complex:·gc.count | thrpt | 25 | 111.000 | counts |
| FleksBenchmark.complex:·gc.time | thrpt | 25 | 60.000 | ms |
| FleksBenchmark.simple | thrpt | 25 | 93.504 ± | 2.781 | ops/s |
| FleksBenchmark.simple:·gc.alloc.rate | thrpt | 25 | 0.296 ± | 0.001 | MB/sec |
| FleksBenchmark.simple:·gc.alloc.rate.norm | thrpt | 25 | 3874.402 ± | 121.604 | B/op |
| FleksBenchmark.simple:·gc.count | thrpt | 25 | ≈ 0 | counts |


## branch with version
| Benchmark | Mode | Cnt | Score | Error | Units |
|-|-|-|-|-|-|
| FleksBenchmark.addRemove | thrpt | 25 | 2548.235 ± | 48.443 | ops/s |
| FleksBenchmark.addRemove:·gc.alloc.rate | thrpt | 25 | 1352.019 ± | 25.684 | MB/sec |
| FleksBenchmark.addRemove:·gc.alloc.rate.norm | thrpt | 25 | 649171.789 ± | 2.297 | B/op |
| FleksBenchmark.addRemove:·gc.churn.G1_Eden_Space | thrpt | 25 | 1359.566 ± | 44.474 | MB/sec |
| FleksBenchmark.addRemove:·gc.churn.G1_Eden_Space.norm | thrpt | 25 | 653034.494 ± | 22035.481 | B/op |
| FleksBenchmark.addRemove:·gc.churn.G1_Survivor_Space | thrpt | 25 | 0.117 ± | 0.026 | MB/sec |
| FleksBenchmark.addRemove:·gc.churn.G1_Survivor_Space.norm | thrpt | 25 | 56.370 ± | 12.668 | B/op |
| FleksBenchmark.addRemove:·gc.count | thrpt | 25 | 201.000 | counts |
| FleksBenchmark.addRemove:·gc.time | thrpt | 25 | 131.000 | ms |
| FleksBenchmark.complex | thrpt | 25 | 1.855 ± | 0.017 | ops/s |
| FleksBenchmark.complex:·gc.alloc.rate | thrpt | 25 | 490.499 ± | 3.819 | MB/sec |
| FleksBenchmark.complex:·gc.alloc.rate.norm | thrpt | 25 | 320261164.053 ± | 1041.862 | B/op |
| FleksBenchmark.complex:·gc.churn.G1_Eden_Space | thrpt | 25 | 488.030 ± | 32.974 | MB/sec |
| FleksBenchmark.complex:·gc.churn.G1_Eden_Space.norm | thrpt | 25 | 318655255.893 ± 21458615.694 | B/op |
| FleksBenchmark.complex:·gc.churn.G1_Survivor_Space | thrpt | 25 | 0.010 ± | 0.006 | MB/sec |
| FleksBenchmark.complex:·gc.churn.G1_Survivor_Space.norm | thrpt | 25 | 6476.000 ± | 3956.521 | B/op |
| FleksBenchmark.complex:·gc.count | thrpt | 25 | 77.000 | counts |
| FleksBenchmark.complex:·gc.time | thrpt | 25 | 55.000 | ms |
| FleksBenchmark.simple | thrpt | 25 | 89.084 ± | 1.233 | ops/s |
| FleksBenchmark.simple:·gc.alloc.rate | thrpt | 25 | 0.294 ± | 0.001 | MB/sec |
| FleksBenchmark.simple:·gc.alloc.rate.norm | thrpt | 25 | 4042.661 ± | 56.375 | B/op |
| FleksBenchmark.simple:·gc.count | thrpt | 25 | ≈ 0 | counts |

If you compare the two it seems that this branch actually requires less garbage collection (gc.count is lower), probably because we no longer need to create Entity classes for iterations.

Quillraven:
Very cool stuff! There was also another issue and PR today (#117) that further reduces GC. I will have a final look over the weekend but once the documentation/comments are ready we can merge it and I will try it out in my three example games that I always take as a reference.

If everything looks good I will release version 2.5 in the near future.

Once more - thank you a lot for your time and help <3

src/commonTest/kotlin/com/github/quillraven/fleks/ComponentTest.kt
dragbone: I recommend not doing that because it would make it too easy to create an invalid implementation of EntityProvider. It is now a requirement to correctly deal with version in any EntityProvider.

dragbone: We could create a method to simplify Entity creation in tests if you want?

src/commonMain/kotlin/com/github/quillraven/fleks/world.kt
dragbone: I created a test to show why I think this is needed: https://github.com/Quillraven/Fleks/pull/115/files#diff-1deae4e923b3764d2c4e62b4862e043ab20d510fbdc80b777dc23fc25117cbfdR825
Currently with version as a Long, it would take 299'865'143 years of continuously recycling the same entity at 1000 frames per second to reach Long.MAX_VALUE. Twice that to wrap around back to version=0. At that point it would start to reuse versions which would only be a problem if the game actually still has a reference to that version and tries to use it.
If we changed version to Int, it would be about 24 days to reach Int.MAX_VALUE or 49 days to reuse version=0.

Given these estimates I think we could go down to storing the version as an Int and still ignore the issue that it is theoretically possible that a version gets reused.

dragbone:
@Quillraven I think I'm done! I added some tests, adjusted comments and answered the last two open questions.
Thank you for your time and being open to the idea of this feature <3

src/commonMain/kotlin/com/github/quillraven/fleks/world.kt
Quillraven: I prefer the `Int` version approach because as you mentioned `Long` is a little bit ridiculous and wastes more memory. Also it is consistent with `id` which is currently also an `Int`.

However, I didn`t get how we solve the overflow issue right now. Doesn`t the version become a negative number at some point and then break everything? I guess maybe not, because the version isn't used for any array index, correct? But still, wouldn't it be nicer to not allow negative versions and simply start over with 0 again?

src/commonTest/kotlin/com/github/quillraven/fleks/ComponentTest.kt
Quillraven: Nah, it is fine, no worries :)

Quillraven:
I thank you for taking your time and implement this feature. It is a great improvement because before, as mentioned in your issue, I always used workarounds for the use case and now we have an elegant solution imo.

Let's change the version to an Int and then I will verify it in my three example games and merge it.

Btw.: Yesterday I tried a few things on your fork. One of them was to remove the compasks of the entity service and put it directly into the entity. I thought this will improve the performance further because the compmask of the entity doesn't need to be looked up anymore. However at least for me it had a negative Impact. I found that interesting and don't understand it to be honest. Maybe you have an explanation for it?

Also, since an entity is now a class we could potentially add a world reference to it and remove the EntityContext concept.

But that is something for the future. And maybe that will also have a negative impact for whatever reason 😅

Anyway, once more, great job and thanks!

src/commonMain/kotlin/com/github/quillraven/fleks/world.kt
dragbone: I remembered that Kotlin has unsigned integers, version is now an UInt ;-)

Quillraven:
Verified it in my example games and everything looks good -> merged

PR #117, opened 2023-09-21
Use inline functions for MutableEntityBag

LobbyDivinus:
Use inline for higher order functions in MutableEntityBag where possible to reduce heap allocations. This only applies when not calling the methods on the interface EntityBag.


Quillraven:
fyi, there is another open PR at the moment (#115 ) which seems to further reduce GC. Takes a little bit more time to merge this feature into Fleks but might be worth for your case to check it out as well.

Thank you for the help - looks good!

PR #121, opened 2023-10-17
version 2.5

Quillraven:
- [x] Changelog
- [x] Wiki update for entity versioning
- [x] Wiki update for entity tags
- [x] Wiki update for serialization + snapshot

----

## Changelog

- NEW (**BREAKING CHANGE**): `Entities` are no longer value classes. They are now a data class that contains the previous `id` and a new `version`. With this new functionality it is now way easier to safely reference entities in components. Big shoutout to @dragbone for the contribution! [#115]
- NEW: A new `EntityTag` functionality was added which is a more memory efficient and faster way to create components without data. Thanks to @LobbyDivinus  for the suggestion and help during development [#119]
- NEW (**BREAKING CHANGE**): a new `Snapshot` class was introduced  to support the new `EntityTag` functionality. Check the wiki's serialization page for more details
- UPDATE: Wiki contains updates for the new functionalities and an updated serialization page using kotlinx-serialization with the latest changes
- UPDATE: improved `ComponentsHolder` memory usage slightly (no longer stores a string property of its name) [#110]
- UPDATE: great performance improvement for `BitArray` usage which makes entity creation and removal even faster. Thank you @tommyettinger for the contribution! [#111]
- UPDATE: memory improvement for `MutableEntityBag`. Thanks to @LobbyDivinus [#117]
- UPDATE: some minor improvements when it comes to serialization with kotlinx-serialization
- UPDATE: following versions got updated:
  - Kotlin 1.9.10
  - Dokka 1.9.10
  - Benchmark 0.4.9
  - Kotlinx-serialization 1.6.0
  - Gradle 8.4
- BUGFIX: fixed a potential issue when iterating over the same family in a nested way while modifying it

### DIFF
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -1,6 +1,6 @@
 # Fleks
 
-[![LTS](https://img.shields.io/badge/LTS-2.4-orange.svg)](https://search.maven.org/artifact/io.github.quillraven.fleks/Fleks/2.4/jar)
+[![LTS](https://img.shields.io/badge/LTS-2.5-orange.svg)](https://search.maven.org/artifact/io.github.quillraven.fleks/Fleks/2.5/jar)
 [![Snapshot](https://img.shields.io/badge/Snapshot-2.5--SNAPSHOT-orange.svg)](https://s01.oss.sonatype.org/#nexus-search;gav~io.github.quillraven.fleks~~2.5-SNAPSHOT~~)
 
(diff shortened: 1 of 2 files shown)


Quillraven:
I will do some final tests on the weekend and will then release the new version. @jobe-m: Might be worth to checkout the new serialization wiki page, its examples and also the new SerializationTest class. I think you can remove your `BaseSerializer` interface in the KorGE example. It should work without it.

Issue #124, opened 2023-11-26
Allow add/remove systems dynamically


metaphore:
While not a strictly critical issue, the absence of system set re-configuration is quite unpleasant. The option to disable systems is good, but sometimes it's preferable to spawn and kill systems entirely. In Ashley, it usually goes hand in hand with `EntitySystem#addedToEngine(Engine)`/`EntitySystem#removedFromEngine(Engine)` and is very convenient for resource management. Systems do not always operate over entities (e.g. `IntervalSystem`).

One good use case could be to toggle debug rendering for a limited time. Here's what it could look like:
```kotlin
class DebugRenderSystem : IteratingSystem(family { ... }) {
    val heavyRenderBatcher = new()
    val heavyInternalResource = new()
    
    // ...
    
    override fun onDispose() {
        super.onDispose()
        heavyRenderBatcher.dispose()
        heavyInternalResource.dispose()
    }
}
```

Systems like this borrow unnecessary resources when they are not enabled and it's easier to just spawn and destroy them when needed. Conditionally creating and accessing such systems during world initialization may look rather dirty too.

This also highlights a little problem of missing a mirrored pair to `onDispose()` (`onInitialize()`?). It's always good to have such mirrored events to acquire/free resources. The system's constructor is not always the best place, as the world is not yet initialized at this point and some required objects may not be available yet (like other systems, that get added after the current one). In general case, it would be good to have that `onInitialize()` method right before the first `onUpdate()` hits. But that of course can be achieved without the library modification.

And lastly, dynamically modifying systems also brings the necessity of order control. Currently, the system set is static, the order is pre-defined, and there's obviously no need for extra sorting. But if systems can be added later, obviously, having the option to put them in the right place is crucial.

Quillraven:
The problem with adding/removing systems dynamically is that it will perform slower than the current solution and you also get some additional tricky situations that can lead to hard to understand bugs. Just to mention a few points while I am against such a change:

- When you add/remove systems during a world `update` call we need to be careful with the systems dataset because usually modifying a dataset while iterating is not allowed and crashes, or does things incorrectly like:
  - Should a system that gets added during iteration still be called? Or skip it until the next update call?
  - Same question with systems that get removed
[...]

PR #127, opened 2023-11-27
#124: add onEnabled/onDisabled functions to IntervalSystem

Quillraven:
PR for #124

@metaphore: Are you fine with that change? I am wondering, if we should call onEnabled/onDisabled also when a system gets created. I mean that we explicitly set it on the constructor (=`init`) block of an IteratingSystem to trigger those functions initially. But I am not 100% sure if this is a good idea. Maybe it is better to really just trigger them, when setEnabled is explicitly called by the user. What do you think?

### DIFF
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/system.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/system.kt
@@ -30,12 +30,22 @@ data class Fixed(val step: Float) : Interval
  */
 abstract class IntervalSystem(
     val interval: Interval = EachFrame,
-    var enabled: Boolean = true,
+    enabled: Boolean = true,
     /**
      * Returns the [world][World] to which this system belongs.
      */
     val world: World = World.CURRENT_WORLD ?: throw FleksWrongConfigurationUsageException()
 ) : EntityComponentContext(world.componentService) {
+
+    var enabled: Boolean = enabled
+        set(value) {
+            if (value) {
+                onEnabled()
+            } else {
+                onDisabled()
+            }
+            field = value
+        }
 
     private var accumulator: Float = 0.0f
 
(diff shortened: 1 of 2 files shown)


metaphore:
Looking good!

> we should call onEnabled/onDisabled also when a system gets created

That sounds good, but in order to make them fully mirrored lifecycle events, that would take also an extra call to `onDisabled` right before the system is destroyed. That gets into the area of the "missing" `initialize()` method that I proposed earlier. But we probably shouldn't go down that road now and can get away with a simple state change listener:

```kotlin
var enabled: Boolean = enabled
    set(value) {
        if (field == value) {
            return
        }
        field = value
        onEnabledChange(value)
    }

/**
* This function gets called whenever the system [enabled] state changes.
*/
open fun onEnabledChange(enabled: Boolean) = Unit
```

Quillraven:
I fixed the issue that the methods got called whene `enabled` didn't change. Also, I kept them as two separate methods because I prefer it that way. In Fleks I always tried to have separate methods for add/remove stuff because there are always cases where you don't need both and I find it annoying to force people to implement a method that isn't needed.

In that specific case I am not a big fan of passing a boolean parameter to such methods because you then end up with:

```kotlin
fun myFun(value: Boolean) {
  if(value) {
    // do that
  } else {
    // do other stuff
  }
}
```

So you are forced to add a nesting level to either execute the "true" path or the "false" path. I guess it is personal preference but I prefer to have separate methods for those two paths and be able to skip one, if not needed at all.

metaphore:
> I prefer to have separate methods for those two paths and be able to skip one, if not needed at all.

I mean there's nothing wrong with that, I agree it's cleaner and more convenient. My only reasoning for `onEnebledChange` was that if `onEnable` and `onDisable` are not called on the system start and `onDispose`, it might be confusing for someone who is not aware of that. But anyway, if you need those to be called just as I explained for a particular system you can do it manually. So lets just keep the PR as is and see if anyone ever complains about it.

Issue #129, opened 2023-11-29
Proposal: call system's onDispose in reverse order.


metaphore:
Strict order is important for both initialization and disposal of the systems and it works nice right now. But I think in many cases it would be better to call `onDispose` in reverse order to keep any possible inter-system relations and resource management in sync within the world.

Honestly, I think it's harder to come up with an example when a forward `onDispose` order is preferred over the reversed. It's usually rather you don't care about `onDispose` order at all or need it to be in reverse. Here's a quick example of a system that uses a resource that is bound to a lifecycle of another system:
```kotlin
class ResourceSystem : IntervalSystem() {

    lateinit var resources: HeavyResource
    
    init {
        resource = loadHeavyResource()
    }
    
    override fun onDispose() {
        resource.dispose
    }
}

class DependentSystem : IntervalSystem() {
    
    init {
        val resource = world.system<ResourceSystem>().resource
        // Need to do something with the resource.
    }
    
    override fun onDispose() {
        val resource = world.system<ResourceSystem>().resource //!! An exception here, as the resource has been disposed at this point.
        // Need to do something with the resource.
    }
}

val world = configureWorld {
    systems {
        add(ResourceSystem()) // Higher order system
        add(DependentSystem()) // Lower order system
    }
}

world.dispose()
```

An argument against this example could be that the systems are all created and destroyed at the same time, and thus their lifecycle is exact. So why keep shared resources within a system, and not in a global context/injection? For me, it's for the sake of encapsulation. I find it often convenient to store and manage some shared objects within specific systems (e.g. post processing shaders and frame buffers, system specific renderers, etc). While global system order makes it very easy to keep track of which resource is initialized/disposed after which.

Going forward with this, when #123 is merged, the same principle of reverse destruction should be applied to the systems's `onRemove` family hooks to keep consistency with the system disposal order.

It takes only a few lines of change to implement, most likely won't break any existing projects, and opens up opportunities for proper object management within systems. Optionally bond to either a system lifetime or an entity lifetime through system family hooks.

Quillraven:
I think you know my point of view about this topic already ;) I suggest to not link systems together as it is against the nature and benefits of ECS.
However, if for whatever reason you need to do it like that and the only change is to call remove listeners/hooks in reverse order I don't mind the change. For me personally it doesn't matter because the order is not important due to modular and independent development of systems.

Feel free to create the PR for it - thanks!

Quillraven:
[...]

PR #130, opened 2023-12-01
#129 reverse system dispose and entity onRemove order.

metaphore:
Follow up for #129

- Call system dispose in reverse order
- Call `FamilyOnRemove.onRemove(entity: Entity)` in reverse order (depends on the bound system).
- Dedicated tests for system execution order.

I'm not sure what's the best way to do a reverse "for" loop on an array in Kotlin. Sounds like there's no built-in singular function for that, and I didn't want to use something that may create an iterator or do unnecessary allocation like `collection.asReversed().forEach()`. So I added an inline extension function `forEachReversed()` at the end of the `world.kt` file. Don't know if there's any better place.

@Quillraven please let me know if there's something wrong with it and if it needs to be changed/moved.


Quillraven:
You are correct that there is no build in function. The only one that I found is:

```kotlin
for (i in array.indices.reversed()) {
    println(array[i])
}
```

According to bytecode it looks like this is a simple for loop without iterator or other objects.

I personally would just use that without a separate extension function but your way is also fine.

PR #133, opened 2023-12-03
Make iterating system private sorting-related fields protected.

metaphore:
Here's a small change that allows for fully custom `IteratingSystem.onTick()` implementations.

It happened so I needed to modify the entity processing flow of the `IteratingSystem` by inserting one extra call right after the sorting is done and before the entity iteration. Like this:
```kotlin
override fun onTick() {
        if (doSort) {
            doSort = sortingType == Automatic
            family.sort(comparator)
        }
        
        // Do something here.

        family.forEach { onTickEntity(it) }
}
```

But it turns out, that I cannot fully replicate the original logic as the sorting-related fields are private.

### DIFF
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/system.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/system.kt
@@ -133,7 +133,7 @@ abstract class IntervalSystem(
  * [Automatic] means that the sorting of [entities][Entity] is happening automatically each time
  * [IteratingSystem.onTick] gets called.
  *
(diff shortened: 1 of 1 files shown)


Quillraven:
What about extracting the sorting logic to a separate `open fun onSort()` method? That way you could more easily add logic before and after sorting. I personally don't like that logic has to be "copy & pasted" when overriding something because it could happen that the original source changes and then you need to be aware of that and also adjust your code.

It is safer to do something like:

```kotlin
override fun onTick() {
  // either add stuff here ...
  super.onSort() // <-- this can also be skipped by users if they never care about sorting in their system
  // ... or here ...
  family.forEach{ onTickEntity(it) }
  // ... or here
}
```

PR #140, opened 2024-04-27
add wasm platform (kotlin -> 1.9.24)

Quillraven:
Trying to add WASM support but seems to be poorly documented and several YouTrack issues opened. From what I know it is still experimental but should be stable with Kotlin 2.0?

Anyway, it is a WIP and I think I resolved all issues except this one:

```
node: bad option: --experimental-wasm-gc
```

From what I understand this flag is no longer needed since it is automatically part of NodeJS? [YouTrack issue](https://youtrack.jetbrains.com/issue/KT-67785/Kotlin-Wasm-Node.JS-22-does-not-need-experimental-wasm-gc-flag-anymore)

So let's wait for Kotlin 2.0 and see if that solves the problem.

In addition, I hope that Kotlin 2.0 also solves following [issue](https://youtrack.jetbrains.com/issue/KT-63014/). We can then remove the part in `kmp-js.gradle.kts`.

Finally, I had to add this in `build.gradle.kts` in root:

```
all {
            // WASM: for bitArray.kt Long::countLeadingZeroBits
            languageSettings.optIn("kotlin.ExperimentalStdlibApi")
        }
```

Maybe that is also not necessary with Kotlin 2.0.

Let's see :)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -27,6 +27,11 @@ kotlin {
     }
 
     sourceSets {
+        all {
+            // WASM: for bitArray.kt Long::countLeadingZeroBits
+            languageSettings.optIn("kotlin.ExperimentalStdlibApi")
+        }
+
         commonMain {
             dependencies {
                 implementation(libs.kotlinxSerialization.json)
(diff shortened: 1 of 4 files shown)


Quillraven:
@soywiz, @jobe-m: I remember that there was an issue/discussion on KorGE if Fleks can be used for wasm. Back then I was not really familiar with the topic, but I think this should now be possible with the Snapshot release that got published a few minutes ago for this PR. There should be now a wasm platform available as well that can be used (thanks to Kotlin 1.9.24 release).

Hope that helps!

jobe-m:
Thanks Simon for the info!
I think this will definitely help if someone want to use it with a warm target. But I don't have any experience with it and even don't know if KorGE has a warm target at all.
In Korge-fleks I am using fleks as source module (compiling it myself through KorGE 's kprojekt) and only for JVM and android currently. So I don't depend on a release from maven or other repo.

PR #148, opened 2024-07-09
System lifecycle

RefuX:
When a system is removed from the world, we should dispose the resources it is using.
If a disposed system is readded to the world it will have to be smart enough to re-create resources or error.


Quillraven:
We also have an onInit function on a system. I am wondering if we should call that when a system gets added? 

Currently it gets called after the world configuration is done.

RefuX:
> We also have an onInit function on a system. I am wondering if we should call that when a system gets added?
> 
> Currently it gets called after the world configuration is done.

Fantastic point, I was wondering about a compliment to onDispose.

Quillraven:
Should be available in SNAPSHOT in a few minutes. Thanks for the contribution!

PR #152, opened 2024-08-29
Be more Kotlin-y and also work with durations.

RefuX:
I was going though my code and removing System timing functions and using the kotlin.time functions instead. 
Though it would be nice for Fleks to support the kotlin.time.Duration value type.

### DIFF
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/world.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/world.kt
@@ -5,6 +5,7 @@ import com.github.quillraven.fleks.collection.MutableEntityBag
 import kotlinx.serialization.Contextual
 import kotlinx.serialization.Serializable
 import kotlin.native.concurrent.ThreadLocal
+import kotlin.time.Duration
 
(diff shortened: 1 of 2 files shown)


Quillraven:
I don't have a problem with that but please add a test to verify that it behaves the same as the float version.

Also, I think delta time in LibGDX is in milliseconds but you use nanoseconds here? Maybe Kotlin has a built in extension function for the conversion that you are doing?


Anyway, I want to release a new version by end of this week. We can add this feature to the release, if everything is ready.

RefuX:
Never used LibGDX, but I'll take your word for it :)

Updated as requested, let me know if you need anything else, don't want to miss the release train. Choo! Choo! 🚂

Quillraven:
@RefuX: I just checked and the delta time in LibGDX is actually in seconds. It is the time between last and current frame in seconds. I will adjust it in the master later and update the documentation of the methods to make it clearer.

Quillraven:
@RefuX : I just pushed the change. Please let me know if this is still fine for you.

Issue #159, opened 2024-10-24
ArrayIndexOutOfBoundsException


AnderWoche:
```kotlin
override fun show() {
        val world = engine.world
        loadScreenSystem(TreasureSystem(world))
        loadScreenSystem(TileMapPathFinderSystem(world))
        loadScreenSystem(TilePathMovementSystem(world))
        loadScreenSystem(AutomaticMoveSystem(world))
        loadScreenSystem(BulletShooterSystem(world))
        loadScreenSystem(EnemyWayWalkerSystem(world)) // crash on this system
    }

    override fun hide() {
        disposeDefenceSystems()
    }

    fun loadScreenSystem(system: IntervalSystem) {
        engine.world.add(system) // it crashed here
        loadedDefenceSystems.add(system)
    }
```

stackTrace:
```kotlin
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
	at com.github.quillraven.fleks.collection.BitArray.hashCode(bitArray.kt:164)
	at com.github.quillraven.fleks.Family.hashCode(family.kt)
	at java.base/java.util.HashMap.hash(HashMap.java:338)
	at java.base/java.util.HashMap.getNode(HashMap.java:568)
	at java.base/java.util.LinkedHashMap.get(LinkedHashMap.java:441)
	at com.github.quillraven.fleks.World.updateAggregatedFamilyHooks(world.kt:506)
	at com.github.quillraven.fleks.World.add(world.kt:232)
	at com.github.quillraven.fleks.World.add(world.kt:242)
	at com.toiletpapertycoon.screen.screens.defence.AbstractDefenceRXScreen.loadScreenSystem(AbstractDefenceRXScreen.kt:31)
```

The System:
```kotlin
class EnemyWayWalkerSystem(world: World) : IteratingSystem(world = world, family = world.family {
    all(EnemyWayWalkerComponent)
}), FamilyOnAdd {

    private lateinit var treasureSystem: TreasureSystem

    override fun onInit() {
        treasureSystem = world.system()
    }

    override fun onAddEntity(entity: Entity) {
// empty funktion and still crashes
    }

```

Quillraven:
Can you provide a simple example on how to reproduce it?

In general: it is not a good practice to link systems together. I see you reference a TreasureSystem and a TileMapPathFinderSystem in your EnemyWayWalkerSystem. That is not a good design. Systems should be separate and ideally don't know from each other. 

[...]

PR #160, opened 2024-10-26
(#159) fixed hashcode function in bitArray

AnderWoche:
We need a Open-ended range

https://kotlinlang.org/docs/ranges.html
```kotlin
// Closed-ended range
println(4 in 1..4)
// true

// Open-ended range
println(4 in 1..<4)
// false
```

### DIFF
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/collection/bitArray.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/collection/bitArray.kt
@@ -160,7 +160,7 @@ class BitArray(
 
         val word = length() ushr 6
         var hash = 0
-        for (i in 0..word) {
+        for (i in 0..< word) {
             hash = 127 * hash + (bits[i] xor (bits[i] ushr 32)).toInt()
         }
         return hash
--- a/src/commonTest/kotlin/com/github/quillraven/fleks/collection/BitArrayTest.kt
+++ b/src/commonTest/kotlin/com/github/quillraven/fleks/collection/BitArrayTest.kt
@@ -1,11 +1,17 @@
 package com.github.quillraven.fleks.collection
 
-import kotlin.test.Test
-import kotlin.test.assertEquals
-import kotlin.test.assertFalse
-import kotlin.test.assertTrue
+import kotlin.test.*
 
 internal class BitArrayTest {
+
+    @Test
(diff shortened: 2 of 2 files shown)


AnderWoche:
I need a snapshot! to continue working!

Quillraven:
Thanks for the PR although I am not 100% sure if the open ended range is correct.

Can you update the test cases to check typical edge cases?

Empty BitArray, first bit last, last bit set, first bit of an additional long set, last bit of an additional long set. Something like that.

You can build Fleks locally. Just execute the publishToMavenLocal task or whatever it is called.

I try to find some time in the evening to have a more detailed look into the issue. If everything looks good I can merge the PR. But I don't want to blindly merge stuff without having a look first.

Quillraven:
@AnderWoche: I had a more detailed look now and while your change will definitely fix the issue, I think it is not the best solution. Why?

The loop is then actually not entered for the first 63 bits of the BitArray resulting in the same hashcode for all of them.

I then also compared once more with the code of LibGDX's Bits class where I think I took the code from. This class actually also crashes in the same scenario :D So it is already broken in LibGDX.

I then checked some different hashCode implementations of other fast BitSet collections in other languages and just replaced the current with a new one.

Also, enhanced the test with some additional test cases and seems good. Please try it out and let me know if it works for you as well!

Good catch and thank you for the help on pointing down the issue!

tommyettinger:
This is a pretty thorny issue, there's a lot of ways something like this can go wrong. `Bits` in libGDX currently (in 1.13.0) can also throw an ArrayIndexOutOfBoundsException in similar cases, I think the code is similar in other places too. I've been experimenting with fixes in [my digital repo](https://github.com/tommyettinger/digital/blob/main/src/test/java/com/badlogic/gdx/utils2/Bits.java#L318-L325); I think this highlighted range should work. The `length() + 63 >>> 6` effectively gets 1/64 of length() rounded up. If the only bit is set with `set(63)`, then that evaluates to 1, and since we use an exclusive range here, then this only hashes bits 0-63, correctly (only long item 0). If we had instead used `set(64)`, then that evaluates to 2, and so this will try to hash what it can of bits 0-127 (long items 0 and 1). If you want me to port that to Kotlin I can, but it should be fairly close to what you already have, just with a different calculation using `length()`, and an exclusive range in the loop.

PR #166, opened 2025-04-13
Replaces Array<T> for List<T> in ComponentsHolder

shaun-wild:
# Changes

✅ All existing tests pass, no new tests required as just changing an implementation detail.

✅ Swaps out Array<T> for List<T>, which removes the need for `reified` types, allowing better generic use of `holder` functions.

# Benchmarks

Performance is critical in Fleks. While this change replaces `Array<T>` with `List<T>`, both are ultimately backed by `Object[]` on the JVM, so data locality and access patterns should remain effectively the same.

I wasn't quite able to benchmark this on my machine as I'm unsure how to compare the numbers (or even if that's possible, as the benchmark results are affected by host machine).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -8,7 +8,7 @@ plugins {
 }
 
 group = "io.github.quillraven.fleks"
-version = "2.12-SNAPSHOT"
+version = "2.13-SNAPSHOT"
 
 kotlin {
     sourceSets {
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/component.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/component.kt
@@ -106,8 +106,10 @@ interface Component<T> {
 class ComponentsHolder<T : Component<*>>(
     private val world: World,
     private val type: ComponentType<*>,
-    private var components: Array<T?>,
+    components: List<T?>,
 ) {
+    private val components = components.toMutableList()
+
     /**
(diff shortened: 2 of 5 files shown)


Quillraven:
Hi,

I am not 100% sure what this is achieving. Why are we removing reified which gives better typesafety during development? Why is a  List better than an Array for generics?

Can you give some examples where you struggled and where this approach helps?

Quillraven:
Here are the benchmarks on my system with and without your changes by running the `FleksOnlyBenchmark` gradle task:

Array:
```
Benchmark                  Mode  Cnt     Score    Error  Units
FleksBenchmark.addRemove  thrpt    5  3846,217 ± 86,121  ops/s
FleksBenchmark.complex    thrpt    5     2,043 ±  0,056  ops/s
FleksBenchmark.simple     thrpt    5   110,848 ±  6,131  ops/s
```

List:
```
jvmBenchmarks summary:
FleksBenchmark.addRemove  thrpt    5  3094,881 ± 172,924  ops/s
FleksBenchmark.complex    thrpt    5     1,893 ±   0,117  ops/s
FleksBenchmark.simple     thrpt    5    97,312 ±  13,187  ops/s
```

So at least on my system this seems to have a rather big impact. I guess this is because of the additional safety things in a List and also the overhead of additional data that it stores. Maybe it also creates some iterator objects at certain areas where before it didn't do that.

PR #167, opened 2025-05-14
change (add system) parameter to generic type for flexibility

befora:
My use case:

```
            systems {
                core.systems.forEach { add(it) }
                game.systems.forEach { add(it) }
            }
```

While iterating system list we need generic type or else `add` function always infers `IntervalSystem`.

### DIFF
--- a/src/commonMain/kotlin/com/github/quillraven/fleks/worldCfg.kt
+++ b/src/commonMain/kotlin/com/github/quillraven/fleks/worldCfg.kt
@@ -57,7 +57,7 @@ class SystemConfiguration(
      *
      * @throws [FleksSystemAlreadyAddedException] if the system was already added before.
      */
-    fun add(system: IntervalSystem) {
+    fun <T: IntervalSystem> add(system: T) {
         if (systems.any { it::class == system::class }) {
             throw FleksSystemAlreadyAddedException(system::class)
         }


befora:
These lists are created lazily so there are some side effects. I think it is ok because systems should not know anything about each other, only their relevant entities.

```
class ExampleSystem : IntervalSystem() {
    val otherSystem = world.system<OtherSystem>() // This is now broken

    override fun onTick() {
        val otherSystem = world.system<OtherSystem>() // This still works
        otherSystem.doSomething()
    }
}
```

Quillraven:
> These lists are created lazily so there are some side effects. I think it is ok because systems should not know anything about each other, only their relevant entities.
> 
> ```
> class ExampleSystem : IntervalSystem() {
>     val otherSystem = world.system<OtherSystem>() // This is now broken
> 
>     override fun onTick() {
>         val otherSystem = world.system<OtherSystem>() // This still works
>         otherSystem.doSomething()
>     }
> }
> ```

I strongly recommend to not doing that. As you said, systems should not know about each other. If you do something like that, I'd say this is a code smell and you should think about a different solution to not couple systems strongly together.

Quillraven:
Change looks fine to me - should be available in the SNAPSHOT version soon. Thanks for the contribution!

PR #171, opened 2025-06-25
use Duration instead of Float

ygdrasil-io:
Related to http://github.com/Quillraven/Fleks/discussions/170

Using **Duration** in over **Float** enhances type safety, readability, and unit clarity. It offers built-in functions, and precision reducing errors.


Quillraven:
Thank you! I try to have a look and run this version against some benchmarks that I have in the next few days.

PR #172, opened 2025-06-26
add coroutines support

ygdrasil-io:


RefuX:
One note about the `SuspendableSystemTest`, call me old fashioned, but I would like to see `SuspendableSystemTestIntervalSystemEachFrame` and `SuspendableSystemTestIntervalSystemFixed` actually having a suspending function in the `onTick` function. A token `delay(1.milliseconds)`, just so it's doing what it says on the tin.

Also IntelliJ (or maybe KtLint) is quite insistent that `override inline suspend fun` is the wrong order and should be `override suspend inline fun.` 🤷‍♂️

Quillraven:
Hey boys, was there any final outcome of this?

Otherwise, I'd like to close the PR for now because I don't like things to be open forever.

PR #173, opened 2025-06-28
abstract delta time to world clock

ygdrasil-io:


Quillraven:
okay the generic world basically changes the entire codebase which is not that great. Is there a simpler approach?

src/commonMain/kotlin/com/github/quillraven/fleks/world.kt
```diff
@@ -26,24 +24,20 @@ fun wildcardSnapshotOf(components: List<Component<*>>, tags: List<UniqueId<*>>):
     return Snapshot(components as List<Component<out Any>>, tags as List<UniqueId<out Any>>)
 }
 
+typealias GenericWorld = World<*>
```
Quillraven, line 27: In general I don't like such alias because it basically hides the generic syntax. When you see `World<Something>` it is obvious that it is a generic type while `GenericWorld` hides that and could be interpreted as something else.

src/commonMain/kotlin/com/github/quillraven/fleks/system.kt
```diff
@@ -28,15 +30,26 @@ data class Fixed(val step: Float) : Interval
...
+    val interval: Interval<T>? = null,
     enabled: Boolean = true,
     /**
      * Returns the [world][World] to which this system belongs.
      */
-    val world: World = World.CURRENT_WORLD ?: throw FleksWrongConfigurationUsageException()
+    val world: GenericWorld = World.CURRENT_WORLD ?: throw FleksWrongConfigurationUsageException()
 ) : EntityComponentContext(world.componentService) {
 
+    private lateinit var _clock: Clock<T>
```
Quillraven, line 42: try to avoid private because there are always people out there who want to overrule something and private just makes it very hard/impossible for them ;)

ygdrasil-io:
> okay the generic world basically changes the entire codebase which is not that great. Is there a simpler approach?

This is the price to pay to get type checking at compile time. We can remove the generics and do unchecked casts, but that would be prone to errors and would not help end users understand how to use the library.

src/commonMain/kotlin/com/github/quillraven/fleks/system.kt
```diff
@@ -28,15 +30,26 @@ data class Fixed(val step: Float) : Interval
  * @param interval the [interval][Interval] in which the system gets updated. Default is [EachFrame].
  * @param enabled defines if the system gets updated when the [world][World] gets updated. Default is true.
  */
-abstract class IntervalSystem(
-    val interval: Interval = EachFrame,
+abstract class IntervalSystem<T>(
+    val interval: Interval<T>? = null,
```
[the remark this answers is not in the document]
ygdrasil-io: EachFrame has been restored, but we cannot keep it as an object since it would mean fixing the generic type, and we can't do that in the constructor parameters.

src/commonMain/kotlin/com/github/quillraven/fleks/world.kt
ygdrasil-io: GenericWorld has been removed

src/commonMain/kotlin/com/github/quillraven/fleks/system.kt
ygdrasil-io: Since onUpdate() is open, is that really necessary? In the base implementation, the clock should not be mutable, and Clock is already an interface that can be extended and set at world creation.

Quillraven:
I just tried it again in a project and it requires quite a lot of code adjustments if we introduce that change. I am not a fan of that and will postpone this big change to a 3.x version which will anyway have bigger breaking changes so it is fine for that version.

However, I don't want you left disappointed! I saw that we actually have some sort of `Duration` functionality already available. I did not remember that haha. It was introduced by @RefuX in 2024. I have a little time now and prepare something. I hope that suits your needs for now.

And really thank you for your time and PRs. Helped me a lot to understand better and judge, if it is going into a good direction or not. I wouldn't have the time right now for such things, unfortunately.

ygdrasil-io:
No problem, that the contract on open source, all proposals cannot be all fulfilled. Do you have an ETA when you are going to start working on v3 ?

Quillraven:
A good question :D Motivation is high but I get my second child next week and kinda busy in private life at the moment. I think (and hope) that I will start end of this year/beginning of next year. So not very soon but it is planned!

Btw, did you have a look at #174? Does that maybe help for now?

PR #174, opened 2025-06-29
proposal for #170 (Duration support)

Quillraven:
@ygdrasil-io: does something simple like that help you for now? It is not finished yet, more like a WIP. The idea is to simply add a `Duration` field to the world that can be accessed by systems. That way we don't break any existing code but you still have the possibility to use the duration object instead of a float.


ygdrasil-io:
> @ygdrasil-io: does something simple like that help you for now? It is not finished yet, more like a WIP. The idea is to simply add a `Duration` field to the world that can be accessed by systems. That way we don't break any existing code but you still have the possibility to use the duration object instead of a float.

I'm going to inject my clock as a parameter into my systems. For v2, this will be sufficient, and I'll make some proposals to add this as an abstraction in v3 :)

PR #179, opened 2025-11-23
Entity Interface

shaun-wild:
# What's going on here

I've been making extensive use of Fleks and I'm trialling creating a multiplayer game, with Fleks at the core. One issue that I have run into: Fleks entities are defined as a `data class`, which means I do not have control over how instances of this class are managed. 

I'd like to use a pooled entity class, as I can avoid creating many Entity instances, for use over the network, which relieves memory/GC pressure in my game.

# Proposed solution

- I have replaced the existing `Entity` data class with an interface and a default `EntityImpl` implementation of said interface.
- I have done a bit of refactoring and hidden the details of this from the user, with the use of `entity` and `emptyEntity` functions.
- Updated the existing usages of Entity to use these new helpers.

# Caveats

- I have not updated documentation; I'm waiting for the green light before doing so. 
- Serialization will work differently when using the `Entity` interface, as you will need to register a polymorphic serializer. 

-----------------------------------------------

Please let me know what you think of this proposed change?


shaun-wild:
The immutable nature of Entities means they do not work well with pooling, meaning I have to create a lot of entity objects each update, which adds GC pressure.

Also no rush, I am not blocked on this.

Quillraven:
But imo an entity must be immutable since it never changes. It is just an ID (+generation counter). Especially when you reference entities in components this is important because you don't want your reference get changed.

Can you elaborate more on your scenario? Why are you creating a lot of entity objects each update? And why does the `EntityProvider` not help in your case?

I am not a big fan of making it an interface + implementation because imo an entity must remain something very simple.

